PR #24, opened 2020-02-19
Fix file closing bug in sample_app. 

pyhys:
Change station name from command line. Print more diagnostic data. Read IP address from Linux kernel.

### DIFF
--- a/sample_app/main_linux.c
+++ b/sample_app/main_linux.c
@@ -5,6 +5,7 @@
 
 #include <arpa/inet.h>
 #include <net/if.h>
+#include <sys/ioctl.h>
 #include <sys/stat.h>
 
 #include <pnet_api.h>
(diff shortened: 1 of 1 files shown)


sample_app/main_linux.c
```diff
@@ -857,6 +828,118 @@ int write_bool_to_file(const char*  filepath, bool value)
...
+
+/**
+ * Copy an IP address (as an integer) to a struct
+ *
+ * @param destination_struct  Out: destination
+ * @param ip                  In: IP address
+*/
+void copy_ip_to_struct(pnet_cfg_ip_addr_t* destination_struct, uint32_t ip)
+{
+   destination_struct->a = (ip&0xFF);
```
hefloryd: nitpicking but please use space between operator and arguments, i.e. (ip & 0xFF)

sample_app/main_linux.c
```diff
@@ -1020,47 +1106,65 @@ int main(int argc, char *argv[])
    printf("\n** Starting Profinet demo application **\n");
    if (verbosity > 0)
    {
-      printf("Verbosity level: %u\n", verbosity);
+      uint8_t macbuffer[6];
+      os_cpy_mac_addr(macbuffer);
```
hefloryd: The profinet stack should not get the mac address like this but rather by configuration, but we can fix that in another PR.

Issue #8, opened 2020-02-04
Protocol violation during IODWrite Request with a single Block


Svenson12:
When establishing a connection, one of the necessary steps is to handle Write Requests. When these requests are done as Multiple Write, everything seems to work fine.
But when such a write request has just one block, the communication fails.
Reason is, that the response contains the right data, but the trasnmitted data length is 0.

Reason for this issue seems to be the last parameter in line 1897 (pf_cmrpc.c). Currently a pointer to the starting position of the response (&res_start_pos) is used which does not make sense. When using p_res_pos instead (simmilar to the multi-write branch), everything seems to work fine...

https://github.com/rtlabs-com/p-net/blob/6ebfff190484a6f61a867f3c757a267b0e63d805/src/device/pf_cmrpc.c#L1897

Can you confirm?

pyhys:
Yes, I can confirm this. Thanks for the detailed bug report! See PR #30

PR #30, opened 2020-03-06
Fix bug in parameter write response.

pyhys:
Closes Github issue #8


pyhys:
Yes, the test will measure the actual response size (for the response that is accepted by the PLC).

PR #37, opened 2020-03-16
Adapt sample app to use multiple modules

pyhys:


hefloryd:
There are no updates to main_rtk.c? It would be great if we could extract common functionality, otherwise both files should be updated.

pyhys:
Updated main_rtk.c so also that one can handle multiple modules.

PR #39, opened 2020-03-26
Read MAC address from hardware

pyhys:


sample_app/main_rtk.c
```diff
@@ -184,9 +186,10 @@ static pnet_cfg_t                pnet_default_cfg =
       /* Network configuration */
       .send_hello = 1,                /* Send HELLO */
       .dhcp_enable = 0,
-      .ip_addr = { 192, 168, 137, 4 },
+      .ip_addr = { 0 },                   /* Read from lwip */
       .ip_mask = { 255, 255, 255, 0 },
```
hefloryd: We can get the netmask from netif_default, in the same way as the other parameters.

PR #41, opened 2020-03-30
New API

pyhys:
Use a network handle, instead of static variables.
Pass a user-defined object to app callbacks.
Add additional status argument to callback indicating data status changes.
Add a generic network handle to osal
Remove Windows files, as there is yet no way to test the new API, and it is unlikely that we ever will fulfill the timing requirements.


src/osal/linux/osal_eth.c
```diff
@@ -27,109 +27,112 @@
...
-
-   timeout.tv_sec =  0;
+   os_eth_handle_t         *handle;
+   int                     i;
+   struct ifreq            ifr;
+   struct sockaddr_ll      sll;
+   int                     ifindex;
+   struct timeval          timeout;
+
+   handle = malloc(sizeof(os_eth_handle_t));
```
hefloryd, line 87: Check that malloc does not return NULL (check all calls of malloc in this commit).

src/osal/rt-kernel/osal.c
```diff
@@ -100,6 +100,8 @@ os_thread_t * os_thread_create (const char * name, int priority,
    return task_spawn (name, entry, priority, stacksize, arg);
 }
 
+/********************** Mutex ************************************************/
```
hefloryd, line 103: Please keep this file similar to other osal.c implementations (in this project and others), to simplify maintenance.

pyhys: Done

src/osal/linux/osal_eth.c
pyhys: Done

PR #45, opened 2020-04-07
Correct implicit read also when there are no slots in use

pyhys:
And handle that AR is null for explicit reads

### DIFF
--- a/src/device/pf_cmsm.c
+++ b/src/device/pf_cmsm.c
@@ -24,6 +24,17 @@
  *
  * The CMSM component monitors the establishment of a connection.
  * Once the device enters the DATA state this component is done.
+ *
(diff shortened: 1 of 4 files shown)


src/device/pf_cmsm.c
```diff
@@ -183,38 +183,40 @@ int pf_cmsm_rm_read_ind(
 {
    int                     ret = -1;
 
-   switch (p_ar->cmsm_state)
+   if (p_ar != NULL)
```
hefloryd, line 186: p_ar != NULL is checked twice in some functions (the second time is always false). Also, in some functions p_ar != NULL is checked after deferencing p_ar. 

Make all functions similar with regards to p_ar if possible, and vaildate parameters and abort early, rather than adding extra indentation levels.

pyhys: Done

PR #48, opened 2020-04-15
Set IP address on Linux

pyhys:
Implement missing functionality in osal.c for Linux.
Use a script for adapting manuipulation to different Linux
distributions.

### DIFF
--- a/cmake/Linux.cmake
+++ b/cmake/Linux.cmake
@@ -61,6 +61,12 @@ target_sources(pn_dev
   sample_app/main_linux.c
   )
 
+file(COPY
+  src/osal/linux/set_network_parameters
+  DESTINATION
+  ${PROFINET_BINARY_DIR}/
+  )
+
 if (BUILD_TESTING)
   set(GOOGLE_TEST_INDIVIDUAL TRUE)
   target_sources(pf_test
--- a/src/device/pnet_api.c
+++ b/src/device/pnet_api.c
@@ -41,6 +41,13 @@ pnet_t* pnet_init(
       return NULL;
    }
 
+   if (strlen(netif) > PNET_MAX_INTERFACE_NAME_LENGTH)
+   {
+      LOG_ERROR(PNET_LOG, "Too long interface name\n");
+      return NULL;
+   }
+   strcpy(net->interface_name, netif);
+
    net->cmdev_initialized = false;  /* TODO How to handle that pf_cmdev_exit() is used before pf_cmdev_init()? */
    net->fspm_log_book_mutex = NULL;  /* TODO is this necessary? */
    net->scheduler_timeout_mutex = NULL;  /* TODO is this necessary? */
--- a/src/osal/linux/osal.c
+++ b/src/osal/linux/osal.c
@@ -27,13 +27,14 @@
 #include <limits.h>
 
 #include <pthread.h>
+#include <sys/syscall.h>
+#include <arpa/inet.h>
 
 #include <assert.h>
 #include <errno.h>
 #include <string.h>
 #include <unistd.h>
 
-#include <sys/syscall.h>
 
 /* Priority of timer callback thread (if USE_SCHED_FIFO is set) */
 #define TIMER_PRIO        5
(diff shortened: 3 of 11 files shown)


src/osal/linux/set_network_parameters
```diff
@@ -0,0 +1,69 @@
...
+   echo "Failed to set default gateway"
+   exit 1
+fi
+
+if ! hostname $HOSTNAME; then
+   echo "Failed to set hostname"
+   exit 1
+fi
+
+exit 0
```
hefloryd, line 69: Add newline at end of file

cmake/Linux.cmake
```diff
@@ -61,6 +61,12 @@ target_sources(pn_dev
   sample_app/main_linux.c
   )
 
+configure_file (
+  src/osal/linux/set_network_parameters
+  ${PROFINET_BINARY_DIR}/
+  COPYONLY
```
hefloryd, line 67: Since replacing contents is unwanted, file(COPY) is maybe clearer and could be useful to specifically set permissions (executable bit).

src/device/pnet_api.c
```diff
@@ -41,6 +41,13 @@ pnet_t* pnet_init(
       return NULL;
    }
 
+   if (strlen(netif) >= PNET_MAX_INTERFACE_NAME_LENGTH)
```
hefloryd, line 44: Is this really correct wrt to overflow and termination? What if netif contains exactly PNET_MAX_INTERFACE_NAME_LENGTH characters?

src/osal/linux/osal.c
```diff
@@ -568,12 +569,61 @@ uint8_t os_buf_header(os_buf_t *p, int16_t header_size_increment)
    return 255;
 }
 
+/**
+ * Convert IPv4 address to string
+ * @param ip               In: IP address
+ * @param outputstring     Out: Outputstring. Should have length INET_ADDRSTRLEN.
+ */
+void ip_to_string(
```
hefloryd, line 577: Since this is not a public function it should have static scope.

src/osal/linux/osal.c
```diff
@@ -568,12 +569,61 @@ uint8_t os_buf_header(os_buf_t *p, int16_t header_size_increment)
...
+   uint32_t                ip,
+   char                    *outputstring)
+{
+   uint32_t                ip_network_endianness = htonl(ip);
```
hefloryd, line 581: I wonder if this would be correct on a big-endian linux? I.e. is the linux representation of an ip address as a 32-bit integer dependent on the endianness of the machine? Not sure. However there are probably many other instances where BE would not work anyway.

Alternatively provide an inverse of the OS_IP4_ADDR_TO_U32 macro and use that?

pyhys: We need `htonl()` for `inet_ntop()` to work. On a big-endian machine `htonl()` would be a no-op, I guess.

It is done pretty much the same for rt-kernel: https://github.com/rtlabs-com/p-net/blob/4a960c8725387a90e511e0975021860d5486f1f2/src/osal/rt-kernel/osal.c#L54

src/osal/linux/set_network_parameters
pyhys: Done

cmake/Linux.cmake
pyhys: Done. The executable bit is set in the git repo, so I'm not modifying it during the copying.

src/device/pnet_api.c
pyhys: Done

src/osal/linux/osal.c
pyhys: Done

src/osal/linux/osal.c
hefloryd: Yes, htonl is needed for a little-endian machine. Whether it is needed for a big-endian machine I'm not sure, it would depend on the internal linux representation of an IP address as a 32-bit integer on a BE target.

As for rt-kernel (LWIP) at a quick check it seems that IP addresses are always in network format, so it would probably be correct there.

However I think this is ok for now but for pnet we should perhaps consider using the platform (network stack) representation of IP address rather than inventing our own and having to convert between representations.

Issue #47, opened 2020-04-15
Invalid Frame when using VLAN IDs unequal 0


Svenson12:
When using VLAN IDs unequal to 0, an invalid frame will be created when sending IO data.
Before line 117 in pf_ppm.c the VLAN type ID should be added to the frame with the following lines:
   u16 = OS_ETHTYPE_VLAN;
   u16 = htons(u16);
   memcpy(&p_payload[pos], &u16, sizeof(u16));
   pos += sizeof(u16);

https://github.com/rtlabs-com/p-net/blob/aa40649996f223369cef28af7cf577a8585a6cb8/src/common/pf_ppm.c#L117

Additionally I would ask, whether it is correct that no VLAN tag is added if the VLAN ID is set to 0. Even if the ID is zero, the tag contains priority information. E.g. the Siemens CPUs send the a VLAN tag with ID 0 and Prio 6. So I would recommend to additionally remove the following conditions to ensure that a VLAN tag is always added:
https://github.com/rtlabs-com/p-net/blob/aa40649996f223369cef28af7cf577a8585a6cb8/src/common/pf_ppm.c#L245
https://github.com/rtlabs-com/p-net/blob/aa40649996f223369cef28af7cf577a8585a6cb8/src/common/pf_ppm.c#L115

pyhys:
Thanks for your feedback, we will take a look!

pyhys:
I have created pull request #49. Please review, @Svenson12

PR #49, opened 2020-04-16
Fix VLAN tag in outgoing cyclic data frames

pyhys:
* Always include the VLAN tag in cyclic data
* Add the VLAN Tag protocol identifier field
* Correct the VLAN priority (it should use the three leftmost bits)

Closes Github issue #47


pyhys:
Thanks for your (as always) very detailed and useful feedback. I have updated the pull request.

PR #59, opened 2020-04-29
Defines for LLDP settings

pyhys:
Add defines for the different LLDP settings, to help end users.

Note that there are a few lower case letters in the defines to increase readability. However it is still obvious that they are defines.

### DIFF
--- a/sample_app/sampleapp_common.c
+++ b/sample_app/sampleapp_common.c
@@ -690,9 +690,11 @@ int app_adjust_stack_configuration(
    stack_config->lldp_cfg.ttl = 20;                /* seconds */
    stack_config->lldp_cfg.rtclass_2_status = 0;
    stack_config->lldp_cfg.rtclass_3_status = 0;
-   stack_config->lldp_cfg.cap_aneg = 3;            /* Supported (0x01) + enabled (0x02) */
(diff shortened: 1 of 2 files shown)


sample_app/sampleapp_common.c
```diff
@@ -690,9 +690,11 @@ int app_adjust_stack_configuration(
    stack_config->lldp_cfg.ttl = 20;                /* seconds */
    stack_config->lldp_cfg.rtclass_2_status = 0;
    stack_config->lldp_cfg.rtclass_3_status = 0;
-   stack_config->lldp_cfg.cap_aneg = 3;            /* Supported (0x01) + enabled (0x02) */
-   stack_config->lldp_cfg.cap_phy = 0x0C00;        /* 100BASE-TX half and full duplex mode capable */
-   stack_config->lldp_cfg.mau_type = 0x0010;       /* Default (copper): 100BaseTXFD */
+   stack_config->lldp_cfg.cap_aneg = PNET_LLDP_AUTONEG_SUPPORTED +
```
hefloryd, line 693: Prefer "|" for bitwise or-operation.

pyhys: Done

PR #60, opened 2020-04-30
Implement more unittests

pyhys:
* Make sure pf_cpm_check_cycle() only returns 0 or -1. This change does not affect existing code, as the only usage check whether the return value is exactly 0.

* Change log level in pf_get_bits()

* Sort entries in test/CMakeLists.txt

* Improve documentation for tested functions.

### DIFF
--- a/src/device/pf_block_reader.c
+++ b/src/device/pf_block_reader.c
@@ -179,12 +179,18 @@ static void pf_get_uuid(
  * @internal
  * Extract bits from a uint32_t.
  * The extracted bits are placed at bit 0 in the returned uint32_t.
+ *
+ *                               A  B  C                      Input. pos = 6, len = 3
+ *  Bit number  ... 12 11 10  9  8  7  6  5  4  3  2  1  0
+ *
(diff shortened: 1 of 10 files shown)


src/device/pf_block_reader.c
```diff
@@ -179,12 +179,18 @@ static void pf_get_uuid(
...
+ *                  ... 0  0  0  0  0  0  0  0  0  A  B  C    Output
+ *
  * @param bits             In:   The bit field.
  * @param pos              In:   First bit to extract.
  * @param len              In:   Number of bits to extract.
- * @return
+ * @return The extracted bits
  */
-static uint32_t pf_get_bits(
+uint32_t pf_get_bits(
```
hefloryd, line 193: Perhaps in another commit; we could probably improve performance considerately by turning the validation of parameters into asserts that would be removed for production builds. Also since it is now public perhaps it should be moved to the header and declared static inline?

pyhys: Great idea!

PR #62, opened 2020-05-05
Implement fragmented DCE RPC responses

### LINKED ISSUE #34: Error when using many parameters
When parameters are declared for sub-modules, the set values are transferred from the PLC to the device during the connection request using write requests. When the size of all parameters to be transferred is bigger than one Ethernet frame, this write request is fragmented.
Receiving these fragmented write requests seems to work, but if the required response becomes bigger than one Ethernet frame, the stack fails. The reason seems to be, that no measure is implemented in pf_cmrpc_dce_packet to handle this situation. I assume the correct reaction would be to send the response fragmented as well, but this is not implemented.
I tried to limit the size by setting "MaxSupportedRecordSize" in the GSDML file to a small value (e.g. 1024), but this violates the GSDML-specification and the GSDML file will be rejected when I try to load it in my S7 project. (I'm using an S7 CPU 1212C with TIA V16.)

### LINKED ISSUE #42: Fails when using many slots
If setting
`#define PNET_MAX_MODULES  17 `
(and modify the GSDML file accordingly) it is still only possible to use 9 input+output modules with the sample app.

When using slot 1-9 in Codesys everything works fine, but when using slot 1-10 this is the output:
```
Parameter write call-back. AREP: 1 API: 0 Slot: 9 Subslot: 1 Index: 123 Sequence: 18 Length: 4
  Bytes: 00 00 00 00 
Parameter write call-back. AREP: 1 API: 0 Slot: 9 Subslot: 1 Index: 124 Sequence: 19 Length: 4
  Bytes: 00 00 00 00 
Parameter write call-back. AREP: 1 API: 0 Slot: 10 Subslot: 1 Index: 123 Sequence: 20 Length: 4
  Bytes: 00 00 00 00 
Parameter write call-back. AREP: 1 API: 0 Slot: 10 Subslot: 1 Index: 124 Sequence: 21 Length: 4
  Bytes: 00 00 00 00 
[ERROR] CMRPC(2498): Unknown packet_type 1
[ERROR] CMRPC(2498): Unknown packet_type 1
Callback on event PNET_EVENT_ABORT. Error class: 253 Error code: 6
Aborting the application
```

pyhys:
This allows larger responses, which is required when using more slots.

Closes #34
Closes #42

### DIFF
--- a/src/device/pf_cmrdr.c
+++ b/src/device/pf_cmrdr.c
@@ -51,7 +51,7 @@ int pf_cmrdr_rm_read_ind(
    uint16_t                start_pos = 0;
    uint8_t                 iocs[255];              /* Max possible array size */
    uint8_t                 iops[255];              /* Max possible array size */
-   uint8_t                 subslot_data[1500];     /* Max possible array size */
+   uint8_t                 subslot_data[PF_FRAME_BUFFER_SIZE];     /* Max possible array size */
    uint8_t                 iocs_len = 0;
    uint8_t                 iops_len = 0;
    uint16_t                data_len = 0;
--- a/src/device/pf_cmrpc.c
+++ b/src/device/pf_cmrpc.c
@@ -175,7 +175,7 @@ void pf_cmrpc_show(
             (unsigned)p_sess->ip_addr & 0xff);
          printf("   port               = %u\n", (unsigned)p_sess->port);
          printf("   sequence_nmb_send  = %u\n", (unsigned)p_sess->sequence_nmb_send);
-         printf("   fragment_nbr       = %u\n", (unsigned)p_sess->fragment_nbr);
+         printf("   fragment_nbr       = %u\n", (unsigned)p_sess->in_fragment_nbr);
          printf("   dcontrol_sequence_nmb = %u\n", (unsigned)p_sess->dcontrol_sequence_nmb);
          printf("   dcontrol result    = %02x %02x %02x %02x\n",
             (unsigned)p_sess->dcontrol_result.pnio_status.error_code,
@@ -924,22 +924,15 @@ static void pf_cmrpc_rm_connect_rsp(
    uint16_t                ix;
    uint16_t                hdr_pos;
    uint16_t                start_pos;
-   uint16_t                res_len = res_size;
 
-   pf_put_pnet_status(p_sess->get_info.is_big_endian, &p_sess->rpc_result.pnio_status, res_len, p_res, p_res_pos);
+   pf_put_pnet_status(p_sess->get_info.is_big_endian, &p_sess->rpc_result.pnio_status, res_size, p_res, p_res_pos);
 
    hdr_pos = *p_res_pos;    /* Save for last */
    /* Insert the response header with dummy length and actual_count */
-   pf_put_uint32(p_sess->get_info.is_big_endian, p_sess->ndr_data.args_length, res_len, p_res, p_res_pos);
-   pf_put_uint32(p_sess->get_info.is_big_endian, p_sess->ndr_data.array.maximum_count, res_len, p_res, p_res_pos);
-   pf_put_uint32(p_sess->get_info.is_big_endian, p_sess->ndr_data.array.offset, res_len, p_res, p_res_pos);
-   pf_put_uint32(p_sess->get_info.is_big_endian, p_sess->ndr_data.array.actual_count, res_len, p_res, p_res_pos);
-
-   /* Limit our response to what the client can handle. */
-   if (*p_res_pos + p_sess->ndr_data.args_maximum < res_size)
-   {
-      res_len = *p_res_pos + p_sess->ndr_data.args_maximum;
-   }
+   pf_put_uint32(p_sess->get_info.is_big_endian, p_sess->ndr_data.args_length, res_size, p_res, p_res_pos);
+   pf_put_uint32(p_sess->get_info.is_big_endian, p_sess->ndr_data.array.maximum_count, res_size, p_res, p_res_pos);
+   pf_put_uint32(p_sess->get_info.is_big_endian, p_sess->ndr_data.array.offset, res_size, p_res, p_res_pos);
+   pf_put_uint32(p_sess->get_info.is_big_endian, p_sess->ndr_data.array.actual_count, res_size, p_res, p_res_pos);
 
    start_pos = *p_res_pos;    /* Start of blocks - save for last */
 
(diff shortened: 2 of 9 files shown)


src/device/pf_cmrdr.c
```diff
@@ -51,7 +51,7 @@ int pf_cmrdr_rm_read_ind(
    uint16_t                start_pos = 0;
    uint8_t                 iocs[255];              /* Max possible array size */
    uint8_t                 iops[255];              /* Max possible array size */
-   uint8_t                 subslot_data[1500];     /* Max possible array size */
+   uint8_t                 subslot_data[PF_FRAME_BUFFER_SIZE];     /* Max possible array size */
```
hefloryd, line 54: For embedded systems, it is generally not a great idea to have huge arrays on the stack. However that can be addressed in another commit.

src/device/pf_cmrpc.c
```diff
@@ -2473,45 +2479,54 @@ static int pf_cmrpc_dce_packet(
...
          case PF_RPC_PT_REQUEST:
+            /* A new request - clear the response buffer */
+            p_sess->out_buf_len = 0;
+            p_sess->out_buf_sent_len = 0;
+            p_sess->out_fragment_nbr = 0;
+
             pf_get_ndr_data(&p_sess->get_info, &req_pos, &p_sess->ndr_data);
             p_sess->get_info.is_big_endian = true;  /* From now on all is big-endian */
 
+            if (max_rsp_len > p_sess->ndr_data.args_maximum)
```
Svenson12, line 2526: In my application this limitation results in a demaged response for the ParameterEnd request.
The payload of the response used here is 32 bytes in this case (I use a S7-1212).
So the complete response frame is limited to 32bytes as well. But it would need 174 bytes, just the payload is 32 bytes.
It seems that the payload and the framesize is mixe here.

With this limitation commented ut all works fined...

src/device/pf_cmrpc.c
```diff
@@ -2536,16 +2552,120 @@ static int pf_cmrpc_dce_packet(
...
+            LOG_DEBUG(PF_RPC_LOG, "CMRPC(%d): Received fragment ACK.\n", __LINE__);
+            /* Handle fragment acknowledgment from the controller. */
+            if (p_sess->out_fragment_nbr > 0)
+            {
+               LOG_DEBUG(PF_RPC_LOG, "CMRPC(%d): Frag: ACK is valid.\n", __LINE__);
+               /* The fragment acknowledgment is valid (expected) */
+               if (p_sess->out_buf_len > p_sess->out_buf_sent_len)
+               {
+                  /* There is more to send - send it (or a part of it) */
+                  if (max_rsp_len > p_sess->ndr_data.args_maximum)
```
Svenson12, line 2596: This code looks similar to the code above. I dont see an actual error, but it should be checked together with the issue above.

Svenson12:
The ParameterEnd request/reply works now. But now the ApplicationReady request is rejected by the S7 with "Fault: seq: 0: status: nca_s_fault_remote_no_memory".
Btw: There is handling for this fault. I did not check it in detail so far, but it seems the session keeps open but should be closed!?

src/device/pf_cmrpc.c
```diff
@@ -2120,49 +2088,54 @@ int pf_cmrpc_rm_ccontrol_req(
...
-      pf_put_uint32(rpc_req.is_big_endian, ndr_data.args_maximum, sizeof(p_sess->buffer), p_sess->buffer, &pos);
-      hdr_pos = pos;
-      pf_put_uint32(rpc_req.is_big_endian, ndr_data.args_length, sizeof(p_sess->buffer), p_sess->buffer, &pos);
-      pf_put_uint32(rpc_req.is_big_endian, ndr_data.array.maximum_count, sizeof(p_sess->buffer), p_sess->buffer, &pos);
-      pf_put_uint32(rpc_req.is_big_endian, ndr_data.array.offset, sizeof(p_sess->buffer), p_sess->buffer, &pos);
-      pf_put_uint32(rpc_req.is_big_endian, ndr_data.array.actual_count, sizeof(p_sess->buffer), p_sess->buffer, &pos);
+      /* Write the _real_ value for args_maximum */
+      ndr_data.args_maximum = PF_FRAME_BUFFER_SIZE - p_sess->out_buf_len;
+      temp_pos = start_pos;
+      pf_put_uint32(rpc_req.is_big_endian, ndr_data.args_maximum, sizeof(p_sess->out_buffer), p_sess->out_buffer, &temp_pos);
```
Svenson12, line 2111: With thsi new block the ApplicationReady request is rejected by the S7 with "Fault: seq: 0: status: nca_s_fault_remote_no_memory". When this block is inactive, everything works fine.
Only difference I can see is, that without this block ArgsMaximum is 1500, with this block it is 1400.
MaximumCount is 1500 in both cases.

pyhys:
Hi again, and thanks for your as always very helpful feedback. I have now updated the pull request slightly.  See PF_CMRPC_MAX_UDP_PAYLOAD_SIZE, which is 1464 bytes. If i send payloads larger than that in an UDP packet, the Linux kernel will itself fragment the frame (and the IOcontroller does not like that).
 
It would be greatly appreciated if you could give it another try.

Svenson12:
Thank you for your prompt reply. The additional change is very valuable but it did not fix the issue.
My proposal is to replace the block I commented on with the following lines:
```
      /* Fixup the header with correct maximum length info. */
      ndr_data.args_maximum = PF_FRAME_BUFFER_SIZE - p_sess->out_buf_len;
      ndr_data.array.maximum_count = PF_FRAME_BUFFER_SIZE - p_sess->out_buf_len;

      /* Write the _real_ value for args_maximum */
      temp_pos = start_pos;
      pf_put_uint32(rpc_req.is_big_endian, ndr_data.args_maximum, sizeof(p_sess->out_buffer), p_sess->out_buffer, &temp_pos);

```
The background is, that the S7 has obviously a problem when args_maximum is smaller than array.maximum_count (which makes sense). So both must be adjusted. There is no need to overwrite array.maximum_count in the output buffer at this place since this is done a couple of lines below.

Svenson12:
And to handle the errors correctly that I observed during my tests, I would recommend to add the following case block in the switch in pf_cmrpc_dce_packet:
```
         case PF_RPC_PT_FAULT:
         case PF_RPC_PT_REJECT:
            pf_session_release(p_sess);
            res_pos = 0;  /* Nothing more to do */
            break;

```
Otherwise the session will never be closed if something is going wrong. And therefor the stack will run out of sessions after some failing communication attempts.
There is also a Cancel reply. Maybe this has to be handled in the same way...

Svenson12:
Looks good now... :-)

pyhys:
Hi again, and thanks for your effort.
We have further simplified the NDR header generation for pf_cmrpc_rm_ccontrol_req().
Float representation in responses is now properly set to IEEE.
Fragmented responses now have correct endianness.
 
If you could take yet another look would be greatly appreciated.

Svenson12:
I tested it again and it still works in my application. So for me it's fine...

pyhys:
Yet some improvements, especially to the session termination.

Sorry to bother you again, but any testing is greatly appreciated.

Svenson12:
I tested it again in my application and it works well.

PR #63, opened 2020-05-15
Split out common functionality for integration tests

pyhys:
* Use a struct for all resulting data from os mocks

* Remove static variables and some of the magic numbers

* Use a struct for callback usage statistics

* Rename file for test utilities


test/test_cmrdr.cpp
```diff
@@ -31,462 +31,34 @@
...
 {
 protected:
-   virtual void SetUp()
+
+   virtual void SetUp() override
    {
       mock_init();
-      init_cfg();
-      counter_reset();
+      cfg_init();
```
hefloryd, line 45: Can we move more common code into the base class? The init functions for instance seems to be called in most/all tests?

pyhys: The idea was to be able to easily adjust the setup for different test cases. So far mock_clear() is really the only one that differ between the different cases. I assume that we would like to test with different configurations in different tests (not yet implemented).

hefloryd: But isn't there a common baseline that all tests would need? Resetting of mock data for instance. A test that has other needs can override the base init if required.

pyhys: Moved more functionality to base class.

PR #65, opened 2020-05-19
Fix potential NULL pointers

pyhys:
Handle malloc failure and resulting memory leak and/or NULL pointer
dereferencing in src/osal/linux/osal.c

Handle possible NULL pointer dereferencing in pf_alarm.c

Found with clang-scan.

### DIFF
--- a/src/common/pf_alarm.c
+++ b/src/common/pf_alarm.c
@@ -915,11 +915,12 @@ static int pf_alarm_apms_a_data_req(
             }
             else if (p_fixed->pdu_type.type == PF_RTA_PDU_TYPE_ERR)
             {
+               CC_ASSERT(p_pnio_status != NULL);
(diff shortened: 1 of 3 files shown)


src/common/pf_alarm.c
```diff
@@ -915,11 +915,18 @@ static int pf_alarm_apms_a_data_req(
             }
             else if (p_fixed->pdu_type.type == PF_RTA_PDU_TYPE_ERR)
             {
-               pf_put_pnet_status(true, p_pnio_status, PF_FRAME_BUFFER_SIZE, p_buf, &pos);
+               if (p_pnio_status != NULL)
```
hefloryd, line 918: Getting here with p_pnio_status is an internal p-net error? If so an assert seems more appropriate.

PR #66, opened 2020-05-20
Move functinality to read network parameters to osal

pyhys:


src/osal.h
```diff
@@ -238,6 +333,26 @@ void os_get_button(
    uint16_t                id,         /* Starting from 0 */
    bool                    *p_pressed);
 
+
```
hefloryd, line 336: These are not really functions that abstract from OS details? The implementation would be the same for all OS:es.

src/osal.h
```diff
@@ -117,6 +121,14 @@ enum os_eth_type {
 typedef uint32_t os_ipaddr_t;
 typedef uint16_t os_ipport_t;
 
+/**
```
hefloryd, line 124: pnet_ethaddr_t should be remove if os_ethaddr_t is added. These types describe the same thing.

src/osal/linux/osal.c
```diff
@@ -15,25 +15,26 @@
...
 #include <pthread.h>
+#include <sys/ioctl.h>
 #include <sys/syscall.h>
-#include <arpa/inet.h>
 
-#include <assert.h>
-#include <errno.h>
-#include <string.h>
-#include <unistd.h>
+#include <osal.h>
```
hefloryd, line 35: It is good practice to include the header that declares the functions provided by a module first in that module. I.e. keep osal.h at the top. This helps ensure that the header is self-contained.

pyhys: Good point. I change to:

-     interface header
-     Headers from same project
-     Linux headers
-     Standard C headers

Sorted in each group

src/osal.h
pyhys: Done

pyhys:
Added convenience functions to print structs defined in osal.h

src/osal/linux/osal.c
```diff
@@ -569,19 +570,33 @@ uint8_t os_buf_header(os_buf_t *p, int16_t header_size_increment)
...
 }
 
-/**
- * Convert IPv4 address to string
- * @param ip               In: IP address
- * @param outputstring     Out: Outputstring. Should have length INET_ADDRSTRLEN.
- */
-static void ip_to_string(
-   uint32_t                ip,
+// TODO Move
```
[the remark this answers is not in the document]
pyhys: Done

src/osal.h
pyhys: Updated osal functions

src/osal.h
hefloryd: Sorry, didn't think this through. With this change pnet_api.h now depends on osal.h, and the installed build doesn't work unless the client also adds osal.h to the include search path. The point of adding pnet_ethaddr_t was to avoid depending on osal.h in the first place.

include/pnet_api.h
```diff
@@ -33,11 +33,13 @@ extern "C"
 {
 #endif
 
-#include <stdint.h>
-#include <stddef.h>
+#include <pnet_export.h>
+#include "osal.h"
```
hefloryd, line 37: See other comment - pnet_api.h should not depend on osal.h.

Either add an internal os_ethaddr_t type or just change the osal mac api:s to use standard types, e.g.:
```

/**
 * Read the MAC address.
 *
 * @param interface_name      In: Name of network interface
 * @param octet1              Out: The first (in transmission order) octet
 * @param octet2              Out: The second octet
 * @param octet3              Out: The third octet
 * @param octet4              Out: The fourth octet
 * @param octet5              Out: The fifth octet
 * @param octet6              Out: The sixth octet
 *
 * @return 0 on success and
 *         -1 if an error occurred
*/
int os_get_macaddress(
   const char              *interface_name,
   uint8_t                 *octet1,
   uint8_t                 *octet2,
   uint8_t                 *octet3,
   uint8_t                 *octet4,
   uint8_t                 *octet5,
   uint8_t                 *octet6
);

```

include/pnet_api.h
```diff
@@ -1067,12 +1069,14 @@ typedef struct pnet_ip_addr_t
...
-typedef struct pnet_ethaddr
-{
-  uint8_t addr[6];
-} pnet_ethaddr_t;
+ * Copy an IP address (as an integer) to a struct
+ *
+ * @param destination_struct  Out: destination
+ * @param ip                  In: IP address
+*/
+PNET_EXPORT void pnet_copy_ip_to_struct(
```
hefloryd, line 1077: Isn't this just a convenience function for the demo? Does it need to be in the stack? See above re avoiding osal.h.

sample_app/main_rtk.c
```diff
@@ -134,22 +142,23 @@ static int _cmd_pnio_run(
...
-   pnet_default_cfg.ip_mask.d = ip4_addr4 (&netif_default->netmask);
-   pnet_default_cfg.ip_gateway.a = pnet_default_cfg.ip_addr.a;
-   pnet_default_cfg.ip_gateway.b = pnet_default_cfg.ip_addr.b;
-   pnet_default_cfg.ip_gateway.c = pnet_default_cfg.ip_addr.c;
-   pnet_default_cfg.ip_gateway.d = 1;
-   memcpy (pnet_default_cfg.eth_addr.addr, netif_default->hwaddr, sizeof(pnet_ethaddr_t));
+   pnet_copy_ip_to_struct(&pnet_default_cfg.ip_addr, ip);
+   pnet_copy_ip_to_struct(&pnet_default_cfg.ip_mask, netmask);
+   pnet_copy_ip_to_struct(&pnet_default_cfg.ip_gateway, gateway);
+   memcpy(pnet_default_cfg.eth_addr.addr, netif_default->hwaddr, sizeof(os_ethaddr_t));
```
hefloryd, line 161: Should use macbuffer?

sample_app/sampleapp_common.c
```diff
@@ -39,6 +39,30 @@ static void print_bytes(uint8_t *bytes, int32_t len)
    printf("\n");
 }
 
+void print_network_details(
```
hefloryd, line 42: This fetches the details again. Isn't the info already available in the stack state? Could the details change between initializing the stack and calling this function?

src/osal/common/osal_helpers.c
```diff
@@ -0,0 +1,42 @@
...
+ * This software is dual-licensed under GPLv3 and a commercial
+ * license. See the file LICENSE.md distributed with this software for
+ * full license information.
+ ********************************************************************/
+
+#include <osal.h>
+
+#include <stdio.h>
+
+ void os_ip_to_string(
```
hefloryd, line 20: Code that doesn't need porting to the OS should obviously not be in the OS porting layer. Move these convenience functions to the demo (or stack if it also needs them?)

pyhys: Done

sample_app/sampleapp_common.c
pyhys: Now passing already known IP address, MAC address etc instead of reading them again.

sample_app/main_rtk.c

include/pnet_api.h
pyhys: Moved to sampleapp

src/osal.h
pyhys: Did put back pnet_ethaddr_t

include/pnet_api.h
pyhys: Did put back pnet_ethaddr_t, and use a os_ethaddr_t

PR #67, opened 2020-05-26
Set valid AR UUID in the release response

pyhys:
Make sure the error class is correct for the release callback.

Found with Automated RT tester.

### DIFF
--- a/src/device/pf_cmrpc.c
+++ b/src/device/pf_cmrpc.c
@@ -1179,7 +1179,7 @@ static int pf_cmrpc_rm_release_interpret_req(
  * @param p_sess           In:   The RPC session instance.
  * @param p_release_io     In:   The release control block.
  * @param res_size         In:   The size of the response buffer.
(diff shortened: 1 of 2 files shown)


src/device/pf_cmrpc.c
```diff
@@ -1958,8 +1963,8 @@ static int pf_cmrpc_rm_write_ind(
  * @param req_pos          In:   Position in the input buffer.
  * @param p_rpc            In:   The RPC header.
  * @param res_size         In:   The size of the response buffer.
- * @param p_res            Out:  The response buffer.
- * @param p_res_pos        InOut:Position in the response buffer.
+ * @param p_res            Out:  The response buffer. Already filled with a DCE/RPC header.
+ * @param p_res_pos        InOut:Position in the response bIuffer.
```
hefloryd, line 1967: spelling

pyhys: Done

PR #70, opened 2020-05-28
Set proper error code for CControl timeout

pyhys:


pyhys:
This relates to #74 but does not fully solves it.

PR #72, opened 2020-05-28
Make generated UUID actually unique

### LINKED ISSUE #69: Include MAC address when creating new UUID
The creation of a new UUID for an outgoing request (like Application Ready) does not create UUIDs that are really unique:
https://github.com/rtlabs-com/p-net/blob/6b83eb308551acd2469c44964de05770c32c18df/src/device/pf_cmrpc.c#L313-L324

As an improvement I would recommend to include at least the MAC address and if possible some timestamp. Additionally the UUID should follow the spec (see https://en.wikipedia.org/wiki/Universally_unique_identifier).

So would recommend something like this:
      p_sess->activity_uuid.data1 = os_get_current_time_us ();
      p_sess->activity_uuid.data2 = net->cmrpc_session_number >> 16;
      p_sess->activity_uuid.data3 = 0x1000 | ((net->cmrpc_session_number >> 8) & 0x00ff);
      p_sess->activity_uuid.data4[0] = 0x80;
      p_sess->activity_uuid.data4[1] = net->cmrpc_session_number & 0xff;
      p_sess->activity_uuid.data4[2] = net->fspm_cfg.eth_addr.addr [0];
      p_sess->activity_uuid.data4[3] = net->fspm_cfg.eth_addr.addr [1];
      p_sess->activity_uuid.data4[4] = net->fspm_cfg.eth_addr.addr [2];
      p_sess->activity_uuid.data4[5] = net->fspm_cfg.eth_addr.addr [3];
      p_sess->activity_uuid.data4[6] = net->fspm_cfg.eth_addr.addr [4];
      p_sess->activity_uuid.data4[7] = net->fspm_cfg.eth_addr.addr [5];
      net->cmrpc_session_number++;

The timestamp does not follow the spec, but the one of my S7 doesn't as well.
With an RTC timestamp you could build an even better UUID, but this could be an issue on some embedded systems.

pyhys:
Closes #69


src/device/pf_cmrpc_helpers.c
```diff
@@ -0,0 +1,55 @@
+#include "pf_includes.h"
```
hefloryd, line 1: includes after header

src/device/pf_cmrpc_helpers.c
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * See RFC 4122
+ *
+ * @param timestamp        In:   Some kind of timestamp
+ * @param session_number   In:   Session number
+ * @param mac_address      In:   MAC address
+ * @param p_uuid           Out:  The generated UUID
+ */
+void pf_generate_uuid(
+   uint32_t                timestamp,
```
hefloryd, line 38: Why not fetch the timestamp in this function?

src/device/pf_cmrpc_helpers.c
pyhys: Done

src/device/pf_cmrpc_helpers.c
pyhys: In order to have a pure function (no side effects and identical output for same input), and to make it easy to unit test it without mocking.

PR #97, opened 2020-06-02
Implement proper error message for invalid DCP GET suboptions

pyhys:


src/common/pf_dcp.c
```diff
@@ -472,6 +483,7 @@ static int pf_dcp_set_req(
    uint8_t                 *p_value)
 {
    int                     ret = -1;
+
    uint8_t                 response_data[3]; /* For negative set */
```
hefloryd, line 487: Use the same name as above (negative_response_data), or is the comment maybe misleading (negative)?

pyhys: Updated the comment.

PR #103, opened 2020-06-04
Implement callback for signal LED state change

### LINKED ISSUE #91: The "Signal" LED does not work from Linux.
Works fine for rt-kernel but needs to be implemented in src/osal/linux/osal.c

Found with Automated RT Tester  case DCP_Signal

pyhys:
Remove LED and button related functionality from OSAL.

Closes #91

### DIFF
--- a/cmake/Linux.cmake
+++ b/cmake/Linux.cmake
@@ -64,6 +64,8 @@ target_sources(pn_dev
 
 file(COPY
   src/osal/linux/set_network_parameters
+  sample_app/set_profinet_leds_linux
+  sample_app/set_profinet_leds_linux.raspberrypi
   DESTINATION
   ${PROFINET_BINARY_DIR}/
   )
--- a/src/osal/rt-kernel/osal.c
+++ b/src/osal/rt-kernel/osal.c
@@ -14,23 +14,19 @@
  ********************************************************************/
 
 #include "osal.h"
-#include "options.h"
+
 #include "log.h"
+#include "options.h"
+
+#include <drivers/net.h>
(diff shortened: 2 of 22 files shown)


src/osal/rt-kernel/osal.c
```diff
@@ -13,24 +13,19 @@
  * full license information.
  ********************************************************************/
 
-#include "osal.h"
-#include "options.h"
 #include "log.h"
+#include "options.h"
+#include "osal.h"
```
hefloryd, line 18: Please keep the module header first in the list of includes (in all files).

pyhys: Yes, sorry.
Done.
Also cmake change to copy the new scripts.

cmake/Linux.cmake
```diff
@@ -68,6 +68,18 @@ file(COPY
   ${PROFINET_BINARY_DIR}/
   )
 
+file(COPY
```
hefloryd, line 71: You can copy multiple files with one file(COPY), as long as the destionation is the same. Also, I think we should use install() here. The main difference between install() and file(COPY) is that the copy takes place at configuration time (when running cmake). If you wanted to install into /usr/local or similar you'd probably need to be root or similar just to run cmake. Whereas with install() you'd only need root for make install.

pyhys: Done. Also adjusted the tests now that #63 have been merged.

pyhys: Changed back to file(COPY) after discussions.

PR #106, opened 2020-06-05
Fix state enums for cmdev ABORT

pyhys:
Use correct enum values in order to make pnet_ar_abort() functional.


Tjohej:
Looks good!

PR #112, opened 2020-06-12
Fix single write

### LINKED ISSUE #111: Devices fails to connect to SIEMENS PLC
When trying to connect to a SIEMENS PLC the Write Response would append the response twice
Bad_Write_Response

To correct this the write response should be written only when the requestor is a doing a WRITE_MULTIPLE.  Otherwise generate response string and append afterwards.  File to be edited is pf_cmrpc.c function pf_cmrpc_rm_write_ind
Bad_Write_Response_Correction

pyhys:
A single write should only have one response block.
For multiple write there should be an initial block signalling
that it is a multiple write, and then the individual response blocks.
    
Closes #111


pyhys:
Screenshot from 2020-06-12 10-35-42

It seems that slot0 subslot0 is repeated after this fix?

PR #117, opened 2020-06-17
Plug DAP module during initialization

olbjo:
This update does not implement correct behavior for the im0_filter_data request but the read result is accepted by the automated RT-tester so test does not stop on im0_filter_data. Related tests still not pass for other reasons.

Modules still defined and plugged by sample application.  Possibly the DAP module shall not be defined by the sample application but this may have other implications.

### DIFF
--- a/sample_app/main_linux.c
+++ b/sample_app/main_linux.c
@@ -662,6 +662,8 @@ int main(int argc, char *argv[])
    appdata_and_stack.appdata = &appdata;
    appdata_and_stack.net = net;
 
+   app_plug_dap(net);
+
    /* Initialize timer and Profinet stack */
    appdata.main_events = os_event_create();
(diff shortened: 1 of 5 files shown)


olbjo:
The comments on the pf_cmdev_init() function says it plugs the DAP module. I can not see that this was is by the stack. 
I thought the slot and subslots for DAP are always the same so I moved them to the pnet_api.h. I need to define them somewhere in the stack to avoid using magic numbers when calling the plug indication callbacks. Maybe define can be moved to gdsml section of file?

One reason I implemented plugging of DAP using the callbacks was that no update of the sample application was needed. If the DAP (sub-)slots shall not be known to the stack  maybe it is better to plug DAP from the sample application after the call to pnet_init(). What do you think?

Tjohej:
Yes. I believe it is better that the sample_application plugs the DAP. It also has the correct ident numbers. The application has to be tailored in each product anyway as it must handle the actual I/O.

PR #118, opened 2020-06-17
Improve scheduling for sending of periodic data

pyhys:
Add documentation how to measure periodicity using Wireshark

### DIFF
--- a/src/common/pf_ppm.c
+++ b/src/common/pf_ppm.c
@@ -19,6 +19,10 @@
  *
  * This handles cyclic sending of data (and IOPS). Initalizes transmit buffers.
  *
+ * One instance of PPM exist per CR (together with a CPM).
+ *
+ * The states are W_START and RUN.
+ *
  * There are functions used by the application (via pnet_api.c) to set and get
  * data, IOCS and IOPS.
  *
@@ -99,8 +103,10 @@ static void pf_ppm_set_state(
  * Insert destination and source MAC addresses, VLAN tag, Ethertype and
  * Profinet frame ID.
  *
+ * Initialize the rest of the buffer to zero.
+ *
  * @param p_ppm            In:   The PPM instance.
- * @param p_buf            InOut:The buffer.
+ * @param p_buf            InOut:The buffer to be initialized.
  * @param frame_id         In:   The frame_id.
  * @param p_header         In:   The VLAN tag header IOCR parameter.
  */
@@ -153,7 +159,7 @@ static void pf_ppm_init_buf(
 
 /**
  * @internal
- * Finalize a PPM transmit message.
+ * Finalize a PPM transmit message in the send buffer.
  *
  * Insert data, cycle counter, data status and transfer status.
  *
@@ -193,12 +199,13 @@ static void pf_ppm_finish_buffer(
  *
  * This is a callback for the scheduler. Arguments should fulfill pf_scheduler_timeout_ftn_t
  *
- * If the PPM has not been stopped during the wait then a data message
+ * If the PPM has not been stopped during the wait, then a data message
  * is sent and the function is rescheduled.
  *
  * @param net              InOut: The p-net stack instance
- * @param arg              In:   The IOCR instance.
- * @param current_time     In:   The current time.
+ * @param arg              In:    The IOCR instance.
+ * @param current_time     In:    The current system time, in microseconds,
+ *                                when the scheduler is started to execute stored tasks.
  */
 static void pf_ppm_send(
    pnet_t                  *net,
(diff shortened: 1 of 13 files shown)


src/common/pf_ppm.c
```diff
@@ -188,42 +194,91 @@ static void pf_ppm_finish_buffer(
...
+ * @param wanted_delay     In:    Wanted delay in microseconds.
+ * @param stack_cycle_time In:    Stack cycle time in microseconds. Must be larger than 0.
+ * @return Number of microseconds of delay to use with the scheduler.
+ */
+uint32_t pf_ppm_calculate_compensated_delay(
+   uint32_t                wanted_delay,
+   uint32_t                stack_cycle_time
+)
+{
+   uint32_t                number_of_stack_tics = 1;  /* We must wait at least one tick */
```
olbjo, line 234: Change "number_of_stack_tics" to "number_of_stack_ticks"

src/common/pf_ppm.c
```diff
@@ -188,42 +194,91 @@ static void pf_ppm_finish_buffer(
...
+
+   if (wanted_delay > stack_cycle_time + stack_cycle_time / 2)
```
olbjo, line 236: Maybe add some extra parentheses: 
if (wanted_delay > (stack_cycle_time + stack_cycle_time / 2))

Svenson12:
Thanks for the good description how to measure the timing.
I was playing a little bit around with the timing. From my point of view there is still an issue. If there is a delay in triggering the PN stack, these delays will be accumulated over the time and the communication will run out of phase.

I tried a solution reusing ppm.exec as described below and it looks good for me. Maybe this can be combined with your solution...

Init p_ppm->exec in the line before pf_scheduler_add in pf_ppm_activate_req:
`      p_ppm->exec = os_get_current_time_us();`

Update ppm.exec in pf_ppm_send right at the beginning (first line in the if block):
`      p_arg->ppm.exec += p_arg->ppm.control_interval;`

Add pass only the remaining delay to pf_scheduler_add in pf_ppm_send:
`         if (pf_scheduler_add(net, p_arg->ppm.exec - os_get_current_time_us(),
                              ppm_sync_name, pf_ppm_send, arg, &p_arg->ppm.ci_timer) == 0)`

The update of ppm.exec at the very end of pf_ppm_send must be removed in this case.

pyhys:
Hi @svenson12, thanks for your input.
I have tested your suggestion on an Ubuntu laptop, and with a reduction ratio of 4 (giving a 4 ms cycle time) there are now and then periods of 3 ms.

I guess the performance is dependent of the hardware capabilities. Which OS and hardware are you using?

Svenson12:
Hi @pyhys, I'm using a customized hardware (32bit CPU with 64MHz core frequency) and an old commercial RTOS providing preemptive multitasking. The PN stack runs together with a "real" application. The application must have a higher prio than the PN stack.

Due to the low core frequency and the (not too high) additional load caused by the application, I limited the cycle time to min 64ms. The stack is triggered all 1ms (if nothing with higher prio is blocking this task).

First I did measurements looking on my S7 to get a reference (shown is min/max/average).
Statistics S7-1212C (min max avg)
You can see there are some pos/neg deviations, but the average interval is almost perfectly 64ms.

Then I tested your improvement in my application:
Statistics with pyhys change (min max avg)
The average interval is ~64.5ms. This is because the pos/neg deviations are quite asymmetric. And this is because there is no compensation of delays caused by blocking of other tasks/interrupts. I assume this may not be visible on a relative high performance system like an laptop.

Finally I tested my proposed improvement:
Statistics with svenson change (min max avg)
There are still relatively significant positive deviations of the interval, but the negative deviations are more or less symmetric. This results is an average interval of ~64.05ms. This should be exactly 64ms, but there is still some offset where I don't know where it comes from. But based on the measurement of the S7 timing I'm somehow sure that I'm measuring correctly.

src/common/pf_ppm.c
pyhys: Done

src/common/pf_ppm.c
pyhys: Done

pyhys:
Updated the algorithm to better take execution variations into account

src/common/pf_scheduler.c
```diff
@@ -402,3 +395,95 @@ void pf_scheduler_show(
...
+uint32_t pf_scheduler_sanitize_delay(
+   uint32_t                wanted_delay,
+   uint32_t                stack_cycle_time,
+   bool                    schedule_half_tick_in_advance
+)
+{
+   uint32_t                number_of_stack_ticks = 1;  /* We must wait at least one tick */
+   uint32_t                resulting_delay = 0;
+
+   CC_ASSERT(stack_cycle_time > 0);
```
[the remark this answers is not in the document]
hefloryd: Nevermind, this is checking for "not 0", not "isn't negative".

pyhys:
Please test @Svenson12, and open an Github issue if you find more problems.

PR #127, opened 2020-06-25
Implement alias name

atomic-3dpenguin:
Added LLDP Implementation so resolve Alias Naming issues


atomic-3dpenguin:
This also addressed issue PeerStationName and PeerPortName

olbjo:
Thanks! I am looking into this.

PR #137, opened 2020-07-01
Ccontrol requests are now re-transmitted up to 3 times if a response …

Tjohej:
…is not received from the controller within 2s.
After that the connection is aborted with the correct error code.


pyhys:
Replaced by #143

PR #154, opened 2020-08-13
Autostart sample application for rt-kernel

pyhys:


sample_app/main_rtk.c
```diff
@@ -264,15 +193,57 @@ int main(void)
...
+   {
+      printf("Failed to initialize p-net application.\n");
+      return -1;
+   }
+
+   app_plug_dap(g_net, gp_appdata);
+
+   if (gp_appdata->arguments.verbosity > 0)
+   {
+      printf("Initialized p-net samople application.\n");
```
hefloryd, line 236: spelling

sample_app/main_rtk.c
```diff
@@ -264,15 +193,57 @@ int main(void)
...
+      printf("Waiting for connect request from IO-controller\n\n");
+   }
+
+   if (gp_appdata->init_done == true)
```
hefloryd, line 240: This can maybe be removed? Probably there to stop the timer if it was already running due to a restart, but that's no longer possible?

sample_app/main_rtk.c

PR #157, opened 2020-08-18
Rework testing

hefloryd:
This pull request reworks the testsuites to fix intermittent segfaults and resource leaks. With these fixes, the testsuites should work on all supported targets.


test/test_dcp.cpp
```diff
@@ -123,37 +123,37 @@ TEST_F (DcpTest, DcpRunTest)
...
    ret = pf_eth_recv(net, p_buf);
    EXPECT_EQ(ret, 1);
 
-   printf("\nGenerating mock flash LED request\n");
+   TEST_TRACE("\nGenerating mock flash LED request\n");
    p_buf = os_buf_alloc(PF_FRAME_BUFFER_SIZE);
    memcpy(p_buf->payload, signal_req, sizeof(signal_req));
    ret = pf_eth_recv(net, p_buf);
    EXPECT_EQ(ret, 1);
    /* 3.5 s should be enough, but on Windows even 5 s is not enough :-( */
```
pyhys, line 155: This comment does not longer apply, even if we build for Windows again (as we do not depend on operating system timers during test)? Also maybe try to reduce to 4 s.
Replace with: 
/* Wait for LED to flash three times at 1 Hz */

hefloryd:
The first check was cancelled because it tries to build from a branch that was accidentally created and then deleted. The second check builds from the correct repo.

src/device/pnet_api.c
```diff
@@ -28,19 +28,13 @@
 #include "pf_includes.h"
 #include "pf_block_reader.h"
 
-pnet_t* pnet_init(
+/* TODO: naming? pnet_init / pnet_create? */
```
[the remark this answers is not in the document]
hefloryd: Done

test/test_dcp.cpp
hefloryd: Done

PR #165, opened 2020-09-03
Perform factory reset and remove data files from sample app

pyhys:


olbjo (approved):

sample_app/main_linux.c
```diff
@@ -30,7 +30,8 @@
 #include <string.h>
 #include <unistd.h>
 
-#define EXIT_CODE_ERROR                1
+#define APP_EXIT_CODE_ERROR            1
```
hefloryd, line 33: There are actually standardized constants for these, EXIT_SUCCESS and EXIT_FAILURE.

pyhys: Nice, done.

PR #166, opened 2020-09-07
Remove printf from stack

pyhys:
Rename printing stack functions to xxx_show()

### DIFF
--- a/doc/implementation_details.rst
+++ b/doc/implementation_details.rst
@@ -51,9 +51,9 @@ Incoming frame PDU types
 * R_Data  RSI
 
 
-Development functions
----------------------
-Use ``pf_show_memory()`` to show the contents of a buffer/memory area.
+Stack development helper functions
+----------------------------------
+Use ``pf_memory_contents_show()`` to show the contents of a buffer/memory area.
 
 
 State machines
--- a/include/pnet_api.h
+++ b/include/pnet_api.h
@@ -1725,6 +1725,16 @@ PNET_EXPORT int pnet_diag_remove(
  *     0x2000              | Show CFG information.
  *     0x4000              | Show scheduler information.
  *     0x8000              | Show I&M data.
+ *
+ *   Bit in the level parameter
+ *   15 14 13 12 11 10 9 8 7 6 5 4 3 2 1 0
+ *    1                                     I&M
+ *       1                                  Scheduler
+ *          1                               CMINA
+ *             1                            AR
+ *                1                         Sessions
+ *                                     1    More IOCR info on AR
+ *                                       1  More IOCR info on AR
  */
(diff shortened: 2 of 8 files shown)


doc/implementation_details.rst
```diff
@@ -42,6 +42,15 @@ The header file ``pnet_export.h`` is generated by cmake. It ends up in
 ``build/include/pnet_export.h``.
 
 
+Stack developent helper functions
```
hefloryd, line 45: Stack development

include/pnet_api.h
```diff
@@ -1725,6 +1725,16 @@ PNET_EXPORT int pnet_diag_remove(
  *     0x2000              | Show CFG information.
  *     0x4000              | Show scheduler information.
  *     0x8000              | Show I&M data.
+ *
```
hefloryd, line 1728: Was this intended? I can't tell from the commit message.

src/device/pf_cmdev.c
```diff
@@ -353,7 +353,7 @@ void pf_cmdev_free_diag(
    }
    else
    {
-      printf("ix =%u\n", (unsigned)item_ix);
+      LOG_ERROR(PNET_LOG, "CMDEV(%d): Failed to free diag with index %u\n", __LINE__, (unsigned)item_ix);
```
hefloryd, line 356: The error is not propagated here. Would an assert be more appropriate?

src/osal/rt-kernel/osal.c
```diff
@@ -412,12 +412,10 @@ os_buf_t * os_buf_alloc(uint16_t length)
...
    return pbuf_alloc(PBUF_RAW, length, PBUF_POOL);
 }
+
 void os_buf_free(os_buf_t *p)
 {
-   if (pbuf_free(p) != 1)
-   {
-      printf("\n\nPBUF not freed\n");
-   }
+   assert(pbuf_free(p) == 1);
```
hefloryd, line 418: Please use ASSERT for rt-kernel (or CC_ASSERT).

src/device/pf_cmdev.c
pyhys: You are right, this is a void xxx_free() function. An out-of-range index should be detected elsewhere (if it is a problem). Here is too late. I will remove the print-out.

src/osal/rt-kernel/osal.c
pyhys: Done

include/pnet_api.h
pyhys: Updated commit message

doc/implementation_details.rst
pyhys: Done

PR #171, opened 2020-09-14
Clean up pf_eth_recv by calculating a frame position

pyhys:
Based on contributions from atomic-3dpenguin


olbjo (approved):
Looks good. 
Buffer length not checked while skipping VLAN tags. This should not be a practical problem.

PR #175, opened 2020-09-16
Free buffer for incoming ethernet frames, at correct location

pyhys:


src/common/pf_alarm.c
```diff
@@ -478,32 +478,29 @@ static int pf_alarm_apmr_high_handler(
...
+      p_apmr_msg->p_buf = p_buf;
+      p_apmr_msg->frame_id_pos = frame_id_pos;
+      if (os_mbox_post(p_apmx->p_alarm_q, (void *)p_apmr_msg, 0) != 0)
+      {
+         LOG_ERROR(PF_ALARM_LOG, "Alarm(%d): Lost one alarm\n", __LINE__);
       }
    }
 
-   return ret;
+   return 1;  /* Means that frame handler should not free buffer, as that will be done when reading the mbox */
```
olbjo, line 503: If alarm is not added to mbox (the error  case), is that a memory leak?  Maybe p_buf shall be freed on mbox_post error?

src/common/pf_alarm.c
```diff
@@ -529,32 +526,28 @@ static int pf_alarm_apmr_low_handler(
...
-            LOG_ERROR(PF_ALARM_LOG, "Alarm(%d): Lost one alarm\n", __LINE__);
-         }
+         p_apmx->apmr_msg_nbr = 0;
+      }
+      p_apmr_msg = &p_apmx->apmr_msg[nbr];
+      p_apmr_msg->p_buf = p_buf;
+      p_apmr_msg->frame_id_pos = frame_id_pos;
+      if (os_mbox_post(p_apmx->p_alarm_q, (void *)p_apmr_msg, 0) != 0)
+      {
+         LOG_ERROR(PF_ALARM_LOG, "Alarm(%d): Lost one alarm\n", __LINE__);
```
olbjo, line 546: p_buf memory leak?

olbjo (commented):
Check out my comments on possible memory leak.

src/common/pf_alarm.c
pyhys: Good point. I have not really modified the code. We need to look into how the mbox handles error cases.

pyhys: Done.

src/common/pf_alarm.c
pyhys: Done also here

olbjo (approved):

PR #180, opened 2020-09-24
Calculate DCP delay in separate function

pyhys:


olbjo (approved):
Looks good. Very quick review of the test code

PR #182, opened 2020-09-24
Improve in-code documentation for DCP

pyhys:


src/common/pf_dcp.c
```diff
@@ -472,10 +496,11 @@ static int pf_dcp_get_req (
  pf_scheduler_timeout_ftn_t
```
olbjo, line 496: insert "*" at start of line

olbjo (approved):

src/common/pf_dcp.c
pyhys: Done

PR #184, opened 2020-09-30
Check that alarm mbox is available before writing to it

pyhys:
Reduce code duplication by using same code for low- and high-
priority alarm frame handlers.

### DIFF
--- a/src/common/pf_alarm.c
+++ b/src/common/pf_alarm.c
@@ -644,44 +644,51 @@ static int pf_alarm_alpmr_apmr_a_data_ind (
    return ret;
 }
 
-/********************** Frame handler callbacks ******************************/
+/*********************** Frame handler callback ******************************/
 
 /**
  * @internal
- * Handle high-priority alarm PDUs from the controller, and puts it into the
+ * Handle an alarm frame from the IO-controller, and put it into the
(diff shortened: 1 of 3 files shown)


src/common/pf_alarm.c
```diff
@@ -748,11 +699,30 @@ static int pf_alarm_apmr_low_handler (
...
+      else
+      {
+        LOG_ERROR (
+           PF_ALARM_LOG,
+           "Alarm(%d): Could not put incoming %s prio alarm frame in"
+               " mbox, as it is deallocated.\n",
+            __LINE__,
+            priotext);
       }
    }
```
fredrikdanielmoller, line 727: Return 1 when p_buf is NULL (as there is no buffer to free)?

fredrikdanielmoller (changes requested):

src/common/pf_alarm.c
pyhys: Good point.
Done.

fredrikdanielmoller (approved):
There is still a race condition here as the two step operation to check mailbox for NULL and posting to it is non-atomic, but this change is an improvement.

olbjo (approved):

PR #186, opened 2020-09-30
Split out more of the common funtionality in the sample application

pyhys:
When pressing button2, cycle between sending alarms, setting
diagnosis and writing to logbook.

### DIFF
--- a/sample_app/main_linux.c
+++ b/sample_app/main_linux.c
@@ -31,8 +31,8 @@
 #include <unistd.h>
 
 #define APP_DEFAULT_ETHERNET_INTERFACE "eth0"
-#define APP_PRIORITY                   15
-#define APP_STACKSIZE                  4096 /* bytes */
+#define APP_THREAD_PRIORITY            15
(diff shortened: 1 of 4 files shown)


sample_app/main_linux.c
```diff
@@ -54,8 +54,9 @@ void show_usage()
            "application LED state.\n");
    printf ("It will also send a counter value (useful also without buttons and "
            "LED).\n");
-   printf ("Button1 value is sent in the periodic data. Button2 triggers an "
-           "alarm.\n");
+   printf ("Button1 value is sent in the periodic data.\n");
+   printf ("Button2 cycles through triggering an alarm, setting diagnosis and "
+           " created logbook entries.\n");
```
olbjo, line 59: "creating" ?

olbjo (approved):
⭐ 
Guldstjärna

sample_app/main_linux.c
pyhys: Done

PR #188, opened 2020-09-30
Save logbook to file

pyhys:
Show logbook and diagnosis details

### DIFF
--- a/src/device/pf_fspm.c
+++ b/src/device/pf_fspm.c
@@ -51,6 +51,38 @@ void pf_fspm_im_show (pnet_t * net)
                                                 for now */
 }
 
+void pf_fspm_logbook_show (pnet_t * net)
+{
+   uint16_t ix;
+
+   printf ("Logbook. Has wrapped: %u\n", net->fspm_log_book.wrap);
+   printf ("ix   Next   Error code            Detail       Timestamp (s)\n");
+
+   for (ix = 0; ix < NELEMENTS (net->fspm_log_book.entries); ix++)
+   {
+      printf ("[%02u] %s      ", ix, (ix == net->fspm_log_book.put) ? "*" : " ");
+
+      if (
+         (net->fspm_log_book.entries[ix].ar_uuid.data1 != 0) ||
+         (net->fspm_log_book.entries[ix].ar_uuid.data2 != 0) ||
+         (net->fspm_log_book.entries[ix].ar_uuid.data3 != 0))
+      {
+         printf (
+            "0x%02X 0x%02X 0x%02X 0x%02X   0x%08" PRIX32 "   %" PRIu32 "\n",
+            net->fspm_log_book.entries[ix].pnio_status.error_code,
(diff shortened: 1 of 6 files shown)


src/device/pf_fspm.c
```diff
@@ -197,6 +229,81 @@ int pf_fspm_validate_configuration (const pnet_cfg_t * p_cfg)
...
+/**
+ * @internal
+ * Save the logbook to nonvolatile memory, if necessary.
+ *
+ * Compares with the content of already stored logbook (in order not to
+ * wear out the flash chip)
+ *
+ * @param net              InOut: The p-net stack instance
+ */
+static void pf_fspm_save_logbook_if_modified (pnet_t * net)
```
olbjo, line 272: I would name the function  "pf_fspm_save_logbook" and skip details on flash wear out on this level. This is handled by the underlying write operation but does not need to be visible in this api.

This is comment only.

src/device/pf_fspm.c
```diff
@@ -365,6 +474,14 @@ void pf_fspm_create_log_book_entry (
...
       net->fspm_log_book.entries[put].entry_detail = entry_detail;
       os_mutex_unlock (net->fspm_log_book_mutex);
+
+      LOG_DEBUG (
+         PNET_LOG,
+         "FSPM(%d): Added logbook entry to position %u\n",
+         __LINE__,
+         put);
+
+      pf_fspm_save_logbook_if_modified (net);
```
olbjo, line 484: It seems like the create function is intended to be used from different contexts since log_book data is protected by the mutex.  However the log_book is not protected during save operation. which may lead to that a faulty log_book is written. Maybe problem can solved by using a temp buffer .  Extending locked block to include save operation is not a good idea

olbjo (changes requested):

src/device/pf_fspm.c
pyhys: Done.

src/device/pf_fspm.c
pyhys: Now using temporary buffers for both read and write.
Done

olbjo:
When I checked "force-pushed" link there are lot of files updated. Is this correct. Original commit had 5 changed files.

hefloryd:
> When I checked "force-pushed" link there are lot of files updated. Is this correct. Original commit had 5 changed files.

Probably correct. Unfortunately the Github review UI sucks compared to say Gerrit. Most likely some minor corrections were made and the PR was rebased against master. All rebased changes now show up in the "force-pushed" link. 

It might help to not rebase PR:s until they are ready to be merged. There is an option to automatically rebase when merging if possible.

src/device/pf_fspm.c
olbjo: ok

src/device/pf_fspm.c
olbjo: ok

olbjo (approved):

PR #194, opened 2020-10-15
Better parsing of LED cyclic data in sample app

pyhys:


sample_app/sampleapp_common.c
```diff
@@ -1214,19 +1217,21 @@ static void app_handle_cyclic_data (
...
       }
-      else
-      {
-         printf ("Wrong outputdata length: %u\n", outputdata_length);
-      }
+      previous_led_state = received_led_state;
+   }
+   else
+   {
+      printf ("Wrong outputdata length: %u or IOPS: %u\n", outputdata_length, outputdata_iops);
```
olbjo, line 1234: I can't see line length in github diff view. However this line is split by the diff viewer. I think this is an indication that it is too long according to coding standard..

olbjo (changes requested):

sample_app/sampleapp_common.c
pyhys: Done. Did run clang-format.

PR #195, opened 2020-10-15
Automatically create zip file of generated documentation

pyhys:


doc/Makefile
```diff
@@ -8,6 +8,8 @@ SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
 
+htmlzipfile := pnet_docs_$(shell date +%Y%m%d).zip
```
hefloryd, line 11: Perhaps use git describe instead (including dirty flag)?

pyhys: Great idea! 
Done.

PR #196, opened 2020-10-16
Add product name to configuration. Add getters for chassis ID

fredrikdanielmoller:
Currently not used. Will be used by SNMP later on.

### DIFF
--- a/src/common/pf_lldp.h
+++ b/src/common/pf_lldp.h
@@ -26,6 +26,49 @@
 extern "C" {
 #endif
 
+#define PF_LLDP_SUBTYPE_MAC              4
+#define PF_LLDP_SUBTYPE_LOCALLY_ASSIGNED 7
+
+/**
+ * Check if information about peer device has been received
+ *
+ * The info is received in an LLDP packet and relevant
+ * info is stored by LLDP stack.
+ *
+ * @param net              In:    The p-net stack instance
+ * @param timestamp_10ms   Out:   Time when the LLDP packet with the info
+ *                                was first received, in units of
+ *                                10 milliseconds with system startup
+ *                                as the zero point (as in SNMP sysUpTime).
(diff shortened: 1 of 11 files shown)


pyhys (approved):

src/common/pf_lldp.h
```diff
@@ -26,6 +26,53 @@
...
+ *
+ * The peer Chassis ID is received in LLDP frames. If no LLDP frame has been
+ * received, the returned Chassis ID is empty.
+ *
+ * Thread-safety is ensured by means of a mutex.
+ *
+ * @param net              In:    The p-net stack instance
+ * @param p_chassis_id     Out:   Chassis ID of peer device.
+ */
+void pf_lldp_get_peer_chassis_id (
```
pyhys, line 72: We need in a few months to support multiple ports (and thus multiple peers).
Would it be possible to use `pnet_lldp_peer_info_t` instead of` pnet_t` as arguments?
This also applies to `pf_lldp_is_peer_info_received()` if we somehow can store the info properly.

It is not entirely clear to me if this function is internal only, or should be used directly by the SNMP implementation. It affects how the arguments should look like.

fredrikdanielmoller: Use of pnet_t* as argument is needed for using the mutex. When multiple port support is added, the function will need to be modified, e.g. by adding a argument for the port number.

The function is not internal to the LLDP module. It will be used by the SNMP implementation.

src/device/pf_block_writer.c
```diff
@@ -3512,6 +3512,8 @@ void pf_put_pdport_data_real (
    /*Input only the number if it's there */
    if (numPeers > 0)
    {
+      pf_lldp_chassis_id_t chassis_id;
```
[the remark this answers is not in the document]
fredrikdanielmoller: Done

PR #201, opened 2020-10-21
Correct return value from pnet_init()

pyhys:
Without this fix the error checking during startup does not work.
This leads to later segfault if the initialisation failed, but
the application not it notified.


hefloryd:
Not a problem introduced by this PR, but resources should preferably be released where they were allocated. The pnet_init_only function should return an error if it fails to initialize, and the caller should free the resource. Note that the testing framework calls pnet_init_only with an instance of pnet_t that was statically allocated, freeing it would be bad.

PR #205, opened 2020-10-22
snmp: Add internal API

fredrikdanielmoller:
Helper functions for use by platform-dependent SNMP server (agent)

The purposes of this header file is to make it easier to implement the glue
code in os_snmp.c by declaring all needed functions in a single header file.
This API should not be used by other modules than the SNMP server.

### DIFF
--- a/src/common/pf_snmp.h
+++ b/src/common/pf_snmp.h
@@ -1,0 +1,801 @@
+/*********************************************************************
+ *        _       _         _
+ *  _ __ | |_  _ | |  __ _ | |__   ___
+ * | '__|| __|(_)| | / _` || '_ \ / __|
+ * | |   | |_  _ | || (_| || |_) |\__ \
+ * |_|    \__|(_)|_| \__,_||_.__/ |___/
+ *
+ * www.rt-labs.com
+ * Copyright 2020 rt-labs AB, Sweden.
+ *
+ * This software is dual-licensed under GPLv3 and a commercial
+ * license. See the file LICENSE.md distributed with this software for
+ * full license information.
+ ********************************************************************/
+
+/**
+ * @file
+ * @brief Helper functions for use by platform-dependent SNMP server (agent).
+ *
+ * The p-net stack does not contain any SNMP server stack. Instead, it relies
+ * on external software to supply the needed SNMP services.
+ * In the platform portability layer (osal), the file os_snmp.c is expected
+ * to implement the glue code needed make the external SNMP server communicate
+ * with the p-net stack (in particular, to get variables stored in the
+ * p-net stack).
+ *
+ * The purposes of this header file is to make it easier to implement the glue
+ * code in os_snmp.c by declaring all needed functions in a single header file.
+ * This API should not be used by other modules than the SNMP server.
+ *
+ * The API assumes the following model:
+ * The Profinet device has a single interface (IP address), but may have
+ * multiple ports. Each port may be connected to 1 or 0 remote network
+ * entities, where a network entity may be a Profinet Device/Controller/
+ * Supervisor or an LLDP-aware network switch.
+ *
+ * Getter functions for the following variables for the local device, interface
+ * and ports are provided:
+ * - sysName
+ * - sysContact
+ * - sysLocation
+ * - sysDescr
+ * - lldpLocChassisId
+ * - lldpLocChassisIdSubtype
+ * - lldpLocPortNum/lldpRemLocalPortNum
+ * - lldpLocPortId
+ * - lldpLocPortIdSubtype
+ * - lldpLocPortDesc
+ * - lldpLocManAddr
+ * - lldpLocManAddrSubtype
+ * - lldpLocManAddrIfId
+ * - lldpLocManAddrIfSubtype
+ * - lldpConfigManAddrPortsTxEnable
+ * - lldpXPnoLocPortNoS
+ * - lldpXPnoLocLPDValue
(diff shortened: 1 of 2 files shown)


src/common/pf_snmp.h
```diff
@@ -0,0 +1,450 @@
...
+ * 
+ * The p-net stack does not contain any SNMP agent stack. Instead, it relies
+ * on external software to supply the needed SNMP services.
+ * In the platform portability layer (osal), the file os_snmp.c is expected
+ * to implement the glue code needed make the external SNMP agent communicate
+ * with the p-net stack (in particular, to get variables stored in the
+ * p-net stack).
+ * 
+ * The purposes of this header file is to make it easier to implement the glue
+ * code in os_snmp.c by delcaring all needed functions in a single header file.
```
olbjo, line 28: "declaring"

olbjo (approved):

pyhys (commented):
This is great!
Most values are now available. I especially like the nice references to the standard.

We need to support multiple local ports, and thus multiple peers.
Maybe use an integer index for port number.
Use some error signalling if the port number not is available.

Related to multiple port some/all of these needs to be implemented:
* lldpLocPortNum (i guess this is the index to use in the API functions for local and peer ports?)
* lldpLocManAddrIfId
* lldpRemLocalPortNum
* lldpRemIndex
* lldpRemManAddrIfId

Some function to ask for uptime is required. The implementation should ask osal for this.

We need to read this value? Should it be the same as ifDesc?
* lldpLocPortDesc

These should be read-write:
* lldpConfigManAddrPortsTxEnable  
* ifAdminStatus
There needs to be setters and getters for these. Should they be stored permanently?

I think it is better if the API has getters and setters for these values, and that it needs to be stored permanently is a detail internal to the stack:
* syslocation
* syscontact
* sysname

What is sysObjectId, and do we need an API function for it?

How much of the info in IfTable needs to be provided by the API? Is everything available automatically in LWIP and in the Linux implementation?

fredrikdanielmoller:
> This is great!
> Most values are now available. I especially like the nice references to the standard.
> 
> We need to support multiple local ports, and thus multiple peers.
> Maybe use an integer index for port number.
> Use some error signalling if the port number not is available.

I will add lldpLocPortNum as a parameter to relevant functions. 

> 
> Related to multiple port some/all of these needs to be implemented:
> 
> * lldpLocPortNum (i guess this is the index to use in the API functions for local and peer ports?)
> * lldpLocManAddrIfId
> * lldpRemLocalPortNum
> * lldpRemIndex
> * lldpRemManAddrIfId
> 
> Some function to ask for uptime is required. The implementation should ask osal for this.

Good suggestion. Will do.

> We need to read this value? Should it be the same as ifDesc?
> 
> * lldpLocPortDesc

Yes, lldpLocPortDesc in LLDP-MIB and ifDesc in MIB-2/IfTable should be the same. I figure this should currently be the "netif" parameter to pnet_init(). Not sure what it should be when multiple ports are supported.

> These should be read-write:
> 
> * lldpConfigManAddrPortsTxEnable
> * ifAdminStatus
>   There needs to be setters and getters for these. Should they be stored permanently?

I'm not convinced these actually need to be writable, at least as used by Profinet.
The "PROFINET Test case specification" says they are writable (8.3.1.2 "IfTable" and 8.3.2.2 "LldpConfigManAddrTable"). However, in the actual test description (8.4.5.1 "Remanence check"), only the objects in 8.3.1.1 ("MIB-II single objects") are mentioned, not the objects in ifTable or LldpConfigManAddrTable.
Also note that changing these value would seriously disrupt the system.
 
> I think it is better if the API has getters and setters for these values, and that it needs to be stored permanently is a detail internal to the stack:
> 
> * syslocation
> * syscontact
> * sysname

Good suggestion. Will do.

> What is sysObjectId, and do we need an API function for it?

sysObjectId tells what kind of device it is. For Profinet, its value is 1.3.6.1.4.1.24686. See PN-AL-Protocol ch.  4.16.5 "Enterprise number for PNIO MIB".
 
> How much of the info in IfTable needs to be provided by the API? Is everything available automatically in LWIP and in the Linux implementation?

Not sure yet. IfTable is provided by lwip, including statistics counters. Writes to ifAdminStatus is disabled by default. Haven't looked at the Linux (net-snmp) implementation of IfTable.

src/common/pf_snmp.h
fredrikdanielmoller: Done

fredrikdanielmoller:
Some changes in last update:
- Port argument added to relevant functions.
- Some variables were added.
- All functions are now getters or setters.
- Setter functions now return errors.
- Getter functions for retrieving information about remote device also return errors.
- Moved some data types to pf_types.h.

pyhys (approved):
Excellent!
It is fine for me to merge this. Just a few considerations:

- Should there also be a function for `sysUpTime`?
- Error information is typically returned as `int` in this stack. Should we use it here as well (instead of `bool`)?
- Which structs should be located in `pf_types.h` @hefloryd ?
- Later on we might need to add interface statistics to this file.

fredrikdanielmoller:
> * Should there also be a function for `sysUpTime`?

I don't think it's needed. The SNMP implementation (i.e. lwip or net-snmp) is expected to have built-in support of SNMP Get or GetNext requests for the sysUpTime variable.

fredrikdanielmoller:
Changed return type to int, with 0 indicating success and -1 failure.

olbjo (approved):
Looks good!

PR #206, opened 2020-10-22
Misc updates to pass test case AR-ASE

olbjo:
When running tests the commits in PR191 Behavior Scenario 2 was also included


pyhys (approved):
Looks great!

PR #210, opened 2020-10-29
CMAKE: Allow p-net to be nested in other CMake project

allannielsen:


pyhys:
Great!
We will have a look at this.

hefloryd:
Thanks for the PR! The reason the include doesn't work for a subproject is that it sets the module path using CMAKE_SOURCE_DIR, which is not correct for a subproject. We should be using CMAKE_CURRENT_SOURCE_DIR instead.

We are working on another PR which will fix this and add some other improvements (like not building tests if p-net is not your main project).

allannielsen:
Happy to hear that this does have your attention.

By the way, I found a work-around. In the parent project I do the following:

    if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/p-net/CMakeLists.txt)
        message(INFO "Building p-net in-line")
        set(HAS_PNET ON)
        set(BUILD_TESTING OFF)
        list(APPEND CMAKE_MODULE_PATH p-net/cmake)
        add_subdirectory(p-net)
    else()
        set(HAS_PNET OFF)
    endif()

Feel free to close the PR if you do not see it fit.

pyhys:
Please see #212

hefloryd:
Your workaround looks similar to what is implemented in #212, so hopefully you won't need any tricks to build p-net now that it is merged. Thanks again.

PR #212, opened 2020-11-03
Use rtlabs-com/osal and rtlabs-com/cmake-tools

hefloryd:
This PR uses the common OSAL and CMake tools repositories to avoid code duplication between rt-labs projects. CMakeLists are updated to facilitate using p-net as a subproject. Tests and documentation are not built when p-net is a subproject.

Remaining p-net-specific abstraction files are renamed to use the pnal prefix, for P-Net Abstraction Layer. Functions and macros have not been renamed in this PR.

Sample-app port-specific code is moved to the corresponding port folder.

The common build tools are consumed as a git submodule. The p-net repository should be cloned with the --recurse-submodules switch. Instructions have been updated accordingly.

Already cloned repositories can be updated with:

`  $ git submodule update --init --recursive
`


hefloryd:
This PR also fixes PR #210.

pyhys (approved):
In README.md:
"make 3.13 or later" => "make 3.14 or later"

PR #216, opened 2020-11-05
Update docs, cmake-tools

hefloryd:


pyhys (approved):
Maybe also add this to `getting_started_linux.rst` and `getting_started_rtkernel.rst`:

```
Already cloned repositories can be updated with::

    git submodule update --init --recursive
```

PR #220, opened 2020-11-10
Remove LLDP TTL from config

pyhys:
The TTL (time to live) value for outgoing LLDP frames should
always be 20 seconds, according to the Profinet standard.
Do not bother users to always supply this value.


olbjo (approved):
I agree. No reason to complicate things.

PR #219, opened 2020-11-06
Add RPC End Point Mapper

olbjo:
Based on contributions in https://github.com/rtlabs-com/p-net/pull/151
by atomic-3dpenguin

-------------------------
Epm check in user scenario 10 passes
Still problems with IP_UPD_RPC_I&M_EPM test case

PR #223, opened 2020-11-10
Behavior scenario 10

olbjo:
Includes epm #219. That one should be merged before this pr


olbjo:
Creating a new PR to avoid messy diffs in review

PR #225, opened 2020-11-11
Support multiple ports in the LLDP configuration

pyhys:


src/common/pf_lldp.c
```diff
@@ -359,17 +357,17 @@ static void pf_lldp_add_ieee_mac_phy (
...
  * @param p_ipaddr         In:    IP address
- * @param p_lldp_cfg       In:    LLDP configuration for this port
+ * @param p_port_cfg       In:    LLDP configuration for this port
  * @param p_buf            InOut: The buffer.
  * @param p_pos            InOut: The position in the buffer.
  */
 static void pf_lldp_add_management (
    const os_ipaddr_t * p_ipaddr,
-   const pnet_lldp_cfg_t * p_lldp_cfg,
+   const pnet_lldp_port_cfg_t * p_port_cfg,
```
fredrikdanielmoller, line 366: This function should not be port-specific. Th Management Address TLV is specific to the interface, not to any physical port connected to the interface.

src/common/pf_lldp.c
```diff
@@ -493,6 +491,19 @@ static void pf_lldp_reset_peer_timeout (pnet_t * net, uint16_t timeout_in_secs)
    }
 }
 
+void pf_lldp_get_port_config (
+   pnet_t * net,
+   int loc_port_num,
+   const pnet_lldp_port_cfg_t ** pp_port_cfg)
+{
+   if (loc_port_num < 1 || loc_port_num >= PNET_MAX_PORT)
```
fredrikdanielmoller, line 499:
```suggestion
   if (loc_port_num < 1 || loc_port_num > PNET_MAX_PORT)
```

src/common/pf_lldp.c
```diff
@@ -548,15 +559,25 @@ void pf_lldp_get_peer_chassis_id (
...
 
-void pf_lldp_send (pnet_t * net)
+/**
+ * Build and send a LLDP message on a specific port.
+ * @param net              InOut: The p-net stack instance
+ * @param loc_port_num     In:    Local port number for port directly
+ *                                connected to the remote device.
+ *                                Valid range: 1 .. N, where N is the total
+ *                                number of local ports used by p-net stack.
+ *                                See pf_snmp_get_port_list().
```
fredrikdanielmoller, line 569:
```suggestion
 * @param loc_port_num     In:    Local port number.
 *                                Valid range: 1 .. N, where N is the total
 *                                number of local ports used by p-net stack.
```

fredrikdanielmoller (changes requested):

src/common/pf_lldp.c
pyhys: I removed this argument. We might later add some interface-related argument if necessary.
Done.

src/common/pf_lldp.c
pyhys: Thanks.
This also shadowed a missing return statement.
Done.

src/common/pf_lldp.c
pyhys: Done

fredrikdanielmoller (approved):

olbjo (approved):

PR #226, opened 2020-11-11
Refactor writing diagnosis to buffer

pyhys:
Breakout witing maint_status to separate function
Correct usage of slot number and subslot number for
filtering and writing.


olbjo (approved):
I can't find any problems.

PR #233, opened 2020-11-13
Send port change notifications

pyhys:
Better summary of available diagnosis when creating an alarm

Based on contributions from atomic-3dpenguin

### DIFF
--- a/src/common/pf_lldp.c
+++ b/src/common/pf_lldp.c
@@ -24,6 +24,8 @@
  * Builds and sends LLDP frames.
  * Receives and parses LLDP frames.
  *
+ * This file should not have any knowledge of Profinet slots, subslots etc.
+ *
  * ToDo: Differentiate between device and port MAC addresses.
  * ToDo: Handle PNET_MAX_PORT ports.
  */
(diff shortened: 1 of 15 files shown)


src/common/pf_lldp.c
```diff
@@ -1020,6 +1020,108 @@ bool pf_lldp_is_peer_info_received (pnet_t * net, uint32_t * timestamp_10ms)
    return is_received;
 }
 
+static void pf_lldp_send_remote_mismatch_alarm (
```
olbjo, line 1023: I think this operation should be placed in pdport which handles the pdport data checks. Operation name could be "pf_pdport_peer_indication(). Operation called when peer info is changed. Nice if lldp do not need to be aware of alarms and mapping of ports to DAP

src/common/pf_lldp.c
```diff
@@ -1020,6 +1020,108 @@ bool pf_lldp_is_peer_info_received (pnet_t * net, uint32_t * timestamp_10ms)
...
+            lldp_peer_info->chassis_id.string,
+            (char *)p_wanted_peer->peer_station_name,
+            lldp_peer_info->chassis_id.len) == 0)
+      {
+         peer_stationname_is_correct = true;
+      }
+
+      // TODO Update also peer_portname_is_correct
+
+      for (ix = 0; ix < NELEMENTS (net->cmrpc_ar); ix++)
```
olbjo, line 1048: The observer checking an alarm rule should not need to care whether there is an active ar or not. I think check should be inside alarm, or the loop will be common to all observers and also application/process alarms?

On the other hand what will happen if we do not have an active ar when the check is performed and the lldp frame is received - then we will never send any alarm ( If I do not misunderstand).

olbjo (changes requested):
Lets discuss!

src/common/pf_lldp.c
pyhys: Done

pyhys:
Replaced by #246

PR #240, opened 2020-11-20
snmp: Change internal API to use iterator

fredrikdanielmoller:
This adds an iterator object which may be used
for iterating over local ports. Previously used
functions are removed.

The rt-kernel port was updated to use the
iterator.


pyhys:
Great!

PR #241, opened 2020-11-20
snmp: Add MIBs for rt-kernel

fredrikdanielmoller:
This adds the two remaining MIBs needed for
passing the conformance test.

Logging was also changed to use osal_log.h,
and the patch file was updated.


pyhys (approved):
Great!

PR #250, opened 2020-11-25
Functions triggering informative callbacks should be void

pyhys:
The return values from some informative user callbacks should not affect the stack.
It is more obvious if the functions triggering them are void.

We could possibly change the public API of the relevant callbacks
later to make them void, but for now use the same return
value meaning for all public callbacks.
Documentation on the public API will be improved in a separate
pullrequest.


olbjo (approved):
We should also check out if the the comments on callbacks in pnet_api.h are correct but I prefer doing that after merging pending PRs which adds some clearifications

PR #251, opened 2020-11-27
Rework alarm

olbjo:
- Peer port check added
- Queue for outgoing alarms added
- Alarms enabled on application ready response from PLC
- PDPort alarms triggered from main context - previously data was not protected properly

With these fixes testcases for alarm and diagnostics pass

### DIFF
--- a/src/pf_types.h
+++ b/src/pf_types.h
@@ -790,6 +790,58 @@ typedef struct pf_alarm_fixed
     */
 } pf_alarm_fixed_t;
 
+typedef struct pf_alarm_err
+{
+   pnet_pnio_status_t pnio_status;
+   /*
+    * pf_alarm_err_t may be followed by alarm_payload:
+    *    vendor_device_error_info = VendorId, DeviceId, data*
+    */
+} pf_alarm_err_t;
+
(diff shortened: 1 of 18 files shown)


src/pf_types.h
```diff
@@ -790,6 +790,58 @@ typedef struct pf_alarm_fixed
...
+   uint16_t next; /* Next in list (array index) */
+} pf_diag_item_t;
+
+#define PF_ALARM_MAX_PAYLOAD_DATA_LEN sizeof (pf_diag_item_t)
+
+typedef struct pf_alarm_payload
+{
+   uint16_t usi;
+   uint16_t len;
+   uint8_t data[PF_ALARM_MAX_PAYLOAD_DATA_LEN];
```
pyhys, line 842: in `pnet_alarm_send_process_alarm()` it says that the max payload length is 1408 bytes. Maybe some compile-time define could be used somehow?

olbjo: To better handle process alarms with large alarm payload, the stack maybe should not copy the data, but require that the application keeps the alarm data until the confirmation (callback) is received?

PR #256, opened 2020-11-30
snmp: Change internal API

fredrikdanielmoller:
Encoding of SNMP variables is now done in the
API functions instead of at the call site.
Some new structs were introduced containing
encoded values.

The following functions now returns new structs:
- pf_snmp_get_management_address()
- pf_snmp_get_link_status()
- pf_snmp_get_peer_management_address()
- pf_snmp_get_peer_link_status()

### DIFF
--- a/src/common/pf_snmp.c
+++ b/src/common/pf_snmp.c
@@ -13,20 +13,109 @@
  * full license information.
  ********************************************************************/
 
-#ifdef UNIT_TEST
-#endif
-
 /**
  * @file
  * @brief Helper functions for use by platform-dependent SNMP server (agent).
  *
  */
 
+#ifdef UNIT_TEST
+#define pf_lldp_get_management_address mock_pf_lldp_get_management_address
+#define pf_lldp_get_peer_management_address                                    \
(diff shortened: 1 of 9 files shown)


src/common/pf_snmp.c
```diff
@@ -13,21 +13,92 @@
...
+#define pf_lldp_get_peer_link_status mock_pf_lldp_get_peer_link_status
+#endif
+
 #include "pf_includes.h"
 #include <string.h>
 
 #define STRINGIFY(s)   STRINGIFIED (s)
 #define STRINGIFIED(s) #s
 
+static uint8_t pf_snmp_bitswapped_byte (const uint8_t byte)
```
pyhys, line 36: I prefer to have documentation on all functions, also the rather trivial ones.

pyhys (approved):

hefloryd (approved):

src/common/pf_snmp.c
fredrikdanielmoller: Done.

Issue #243, opened 2020-11-22
Remove private types from public header file


fredrikdanielmoller:
Currently, there are several types defined in `pnet_api.h` which are not actually used by any public API function. Examples: `pnet_lldp_peer_info_t` and `pf_lldp_chassis_id_t`.

Move those to a private header file, such as `pf_types.h`, as to minimise the API surface.

pyhys:
- pnet_lldp_peer_info_t
- pf_lldp_chassis_id_t
- pnet_im_nvm_t
- pnet_lldp_peer_to_peer_boundary_t
- pnet_lldp_boundary_t
- PNET_DIAG_CH_XXX constants and macros

Anything else?

fredrikdanielmoller:
* PNET_FILENAME_*
* pnet_profibus_delay_t
* pnet_ieee_macphy_t

PR #257, opened 2020-12-02
Reduce size of public API

fredrikdanielmoller:
Some structs and #defined constants were defined
in the public header file but not used for any
public API function.

By moving them to private header and source files,
the public API is made smaller and thus easier to
use.

Compile-time checks for the FILENAME
constants were also added.

See #243.


src/device/pf_fspm.c
```diff
@@ -33,6 +33,14 @@
 #include "pf_includes.h"
 #include "pf_block_reader.h"
 
+typedef struct pnet_im_nvm
```
pyhys, line 36: pf_im_nvm

src/pf_types.h
```diff
@@ -2323,6 +2323,13 @@ typedef struct pf_interface_stats
    uint32_t if_out_errors;
 } pnet_interface_stats_t;
 
+typedef struct pnet_ieee_macphy_config
+{
+   uint8_t aneg_support_status; /**< Autonegotiation status */
+   uint16_t aneg_cap;           /**< Autonegotiation capabilites */
+   uint16_t operational_mau_type;
+} pnet_ieee_macphy_t;
```
pyhys, line 2331: pf_ieee_macphy_t;

pyhys (commented):
Great!
However I think the structs we move from pnet_api.h should be renamed pf_xxx.
See a few examples below.

src/device/pf_fspm.c
fredrikdanielmoller: Done (in separate Commit).

src/pf_types.h
fredrikdanielmoller: Done (in separate Commit).

fredrikdanielmoller:
Rebased. Changed prefix pnet_ -> pf_ in separate Commit.

src/pf_types.h
```diff
@@ -2464,6 +2495,46 @@ typedef struct pf_lldp_signal_delay
...
+typedef struct pf_lldp_peer_to_peer_boundary
+{
+   pf_lldp_boundary_t boundary;
+   uint16_t properites;
+} pf_lldp_peer_to_peer_boundary_t;
+
+/**
+ * LLDP Peer information used by the Profinet stack.
+ */
+typedef struct pft_lldp_peer_info
```
pyhys, line 2517: `pf_lldp_peer_info`

fredrikdanielmoller: Done.

pyhys (approved):
Great work!
This helps the end users by reducing the clutter.

fredrikdanielmoller:
Rebase

olbjo (approved):
Great!

PR #263, opened 2020-12-04
Rename pnet_set_state() to pnet_set_primary_state()

pyhys:
It is difficult to know what a function pnet_set_state() is about.
With the new name it is easier to understand that it is related
to primary and backup state of the cyclic data communication.


olbjo (approved):
👍

PR #266, opened 2020-12-05
Constify parsed buffer

fredrikdanielmoller:
Adds the 'const' qualifier on the buffer in the
type pf_get_info_t used for parsing a received
packet.


pyhys (approved):
Sweet!

olbjo (approved):

PR #268, opened 2020-12-08
Fix faulty padding for pdport data real and data check

olbjo:
I will mark this ready when I have rerun all automated rt-tests.


pyhys (approved):
Great!

PR #269, opened 2020-12-08
Adjust sampleapp systemd file to new installation path

pyhys:


sample_app/pnet-sampleapp.service
```diff
@@ -3,15 +3,18 @@
...
 [Service]
 Type=simple
+ExecStartPre=/sbin/ifconfig eth0 up
 ExecStartPre=/bin/sh -c '/bin/echo "49152 60999" > /proc/sys/net/ipv4/ip_local_port_range'
 ExecStartPre=/usr/bin/gpio export 22 in
 ExecStartPre=/usr/bin/gpio export 27 in
-WorkingDirectory=/home/pi/profinet/p-net/build/
-ExecStart=/home/pi/profinet/p-net/build/pn_dev -v -b /sys/class/gpio/gpio22/value -d /sys/class/gpio/gpio27/value
+WorkingDirectory=/home/pi/profinet/build/
+ExecStart=/home/pi/profinet/build/pn_dev -v -b /sys/class/gpio/gpio22/value -d /sys/class/gpio/gpio27/value
```
olbjo, line 15: Maybe default should be to run without the real gpios?
Only with -v option. gpio version can be kept but commented out but available for advanced users.
This version works for my build/setup.

olbjo (approved):
Great!

sample_app/pnet-sampleapp.service
pyhys: I do not think using inputs is affecting anything negatively?

olbjo: 👍

PR #270, opened 2020-12-09
Fix DCP SAM

olbjo:
Problem detected in test case DCP Access. Problem was that
timeouts were not removed but only added giving a random behavior.

### DIFF
--- a/src/common/pf_dcp.c
+++ b/src/common/pf_dcp.c
@@ -28,7 +28,7 @@
  * Registers the different frame IDs for these messages with the frame handler.
  *
  * Responses to IDENTIFY are delayed (by an amount calculated from the
- * MAC address), as many devices might resond to the same request.
+ * MAC address), as many devices might respond to the same request.
  *
  * The SAM (Source Address ?) makes sure that that just a single remote
  * MAC address is used for the communication.
@@ -65,6 +65,10 @@
 #define PF_DCP_GET_SET_FRAME_ID 0xfefd
 #define PF_DCP_ID_REQ_FRAME_ID  0xfefe
 #define PF_DCP_ID_RES_FRAME_ID  0xfeff
+
+/* Client hold time, 3 seconds.
+   See Profinet 2.4 Services, section 6.3.11.2.2 */
+#define PF_DCP_SAM_TIMEOUT 3000000 /* microseconds */
 
 CC_PACKED_BEGIN
 typedef struct CC_PACKED pf_ethhdr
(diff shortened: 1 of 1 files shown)


src/common/pf_dcp.c
```diff
@@ -204,9 +205,57 @@ static void pf_dcp_responder (pnet_t * net, void * arg, uint32_t current_time)
...
+/**
+ * @internal
+ * Reset SAM timeout
+ * SAM limits the communication to a single remote MAC address
+ * for 3 seconds.
+ *
+ * @param net              InOut: The p-net stack instance
+ * @param mac              In: Remote MAC address
+ */
+static void pf_dcp_reset_sam_timeout (pnet_t * net, const pnet_ethaddr_t * mac)
```
pyhys, line 221: Maybe `_restart_ `instead to explain that the timer is started, not stopped.

src/common/pf_dcp.c
```diff
@@ -204,9 +205,57 @@ static void pf_dcp_responder (pnet_t * net, void * arg, uint32_t current_time)
...
+      PF_DCP_LOG,
+      "DCP(%d): Update SAM mac and reset timeout\n",
+      __LINE__);
+
+   memcpy (&net->dcp_sam, mac, sizeof (net->dcp_sam));
+   (void)pf_scheduler_remove (net, dcp_sync_name, net->dcp_sam_timeout);
+   net->dcp_sam_timeout = 0;
+   (void)pf_scheduler_add (
+      net,
+      3 * 1000 * 1000, /* 3s */
```
pyhys, line 235: I think we should have a define for this.

pyhys (commented):
Great!
Some minimal adjustment.

src/common/pf_dcp.c
olbjo: Done

src/common/pf_dcp.c
olbjo: Done

pyhys (approved):
Great!

src/common/pf_dcp.c
```diff
@@ -66,6 +66,8 @@
 #define PF_DCP_ID_REQ_FRAME_ID  0xfefe
 #define PF_DCP_ID_RES_FRAME_ID  0xfeff
 
+#define PF_DCP_SAM_TIMEOUT_3_SEC 3000000
```
hefloryd, line 69: Is this is a define that could be set to something other than 3s (could it be shorter maybe)? If so change the name so that it doesn't contain the value.

pyhys: Maybe:
```
/* Client hold time, 3 seconds.
   See Profinet 2.4 Services, section 6.3.11.2.2 */
#define PF_DCP_SAM_TIMEOUT 3000000 /* microseconds */
```

PR #272, opened 2020-12-10
Multi pdport

olbjo:
Almost completes mapping of multiple ports for pdport config files and profinet index read / write. 
What still remains is to handle interface vs ports.


pyhys (commented):
If I search for PNET_SUBMOD and PNET_SUBSLOT I'm also finding stuff in:
```
src/device/pf_block_writer.c
src/device/pf_cmrdr.c
src/common/pf_alarm.c
```
It would great if we also could fix `pf_port_get_list_of_ports()`

src/device/pf_pdport.c
```diff
@@ -154,21 +187,40 @@ static int pf_pdport_save (pnet_t * net)
...
 static uint16_t pf_pdport_loc_port_num_to_dap_subslot (int loc_port_num)
 {
-   CC_ASSERT (loc_port_num == PNET_PORT_1);
-   return PNET_SUBSLOT_DAP_INTERFACE_1_PORT_1_IDENT;
+   uint16_t subslot =
+      PNET_SUBSLOT_DAP_INTERFACE_1_PORT_1_IDENT + PNET_PORT_1 - loc_port_num;
+   return subslot;
+}
+
+static int pf_pdport_dap_subslot_to_local_port (uint16_t subslot)
```
pyhys, line 204: Add docstring.
Describe that 0 is returned for subslots out of range.

src/device/pf_pdport.c
```diff
@@ -154,21 +187,40 @@ static int pf_pdport_save (pnet_t * net)
...
+{
+   int port = PNET_PORT_1 + PNET_SUBSLOT_DAP_INTERFACE_1_PORT_1_IDENT - subslot;
+   if (port < PNET_PORT_1 || port > PNET_MAX_PORT)
+   {
+      port = 0;
+   }
+   return port;
+}
+
+static bool pf_pdport_subslot_is_dap_port_id (uint16_t subslot)
```
pyhys, line 214: Add docstring

src/device/pf_pdport.c
```diff
@@ -228,69 +283,92 @@ static void pf_pdport_remove_all_diag (pnet_t * net)
...
 
-void pf_pdport_ar_connected (pnet_t * net, const pf_ar_t * p_ar)
+void pf_pdport_remove_data_files (const char * file_directory)
 {
-   /* Todo check which dap slots are active in ar and map to ports */
-
-   pf_port_t * p_port_data = pf_port_get_state (net, PNET_PORT_1);
-   if (p_port_data->pdport.adjust.active == true)
+   int port;
+   for (port = PNET_PORT_1; port <= PNET_MAX_PORT; port++)
```
pyhys, line 330: Please use `pf_port_get_next()`

src/device/pf_pdport.c
```diff
@@ -228,69 +283,92 @@ static void pf_pdport_remove_all_diag (pnet_t * net)
...
+   port = pf_port_get_next (&port_iterator);
 
-   pf_port_t * p_port_data = pf_port_get_state (net, PNET_PORT_1);
-   if (p_port_data->pdport.adjust.active == true)
+   while (port != 0)
    {
-      if (
-         p_port_data->pdport.adjust.peer_to_peer_boundary.peer_to_peer_boundary
-            .do_not_send_LLDP_frames == 1)
+      if (p_port_data->pdport.adjust.active == true)
```
pyhys, line 353: We need to do `pf_port_get_state() `first?

pyhys (commented):
Great!
Just some minor details.

src/device/pf_pdport.c
olbjo: Yes, that was causing the unit test failure.
Done

src/device/pf_pdport.c
```diff
@@ -368,11 +464,11 @@ int pf_pdport_read_ind (
       }
       break;
    case PF_IDX_DEV_PDREAL_DATA:
-      if (
-         (p_read_req->slot_number == PNET_SLOT_DAP_IDENT) &&
-         ((p_read_req->subslot_number == PNET_SLOT_DAP_IDENT) ||
-          (p_read_req->subslot_number == PNET_SUBMOD_DAP_IDENT)))
+      if (((subslot == PNET_SUBSLOT_DAP_WHOLE_MODULE) ||
+           (subslot == PNET_SUBMOD_DAP_IDENT)))
```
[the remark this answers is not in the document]
olbjo: Done

src/device/pf_pdport.c
olbjo: This is only used in a special case where calling function do not have "net". 
"net" is a unused parameter in the port iterator interface so I could remove it there but maybe there is an idea that "net" may be useful in the future in the port iterator. I did not dare to remove or starting to add "net" in used callchain.

src/device/pf_pdport.c
olbjo: Done

src/device/pf_pdport.c
olbjo: Done

src/device/pf_pdport.c
pyhys: Fine! Maybe add:
`/* Do not use port iterator as net not is available */`

src/common/pf_alarm.c
```diff
@@ -2731,19 +2731,45 @@ int pf_alarm_send_diagnosis (
...
-      // device ‘b’
+      ret = pf_cmdev_get_module_ident (net, api_id, slot_nbr, &module_ident);
+      if (ret != 0)
+      {
+         LOG_ERROR (
+            PF_ALARM_LOG,
+            "Alarm(%d): Failed to get module ident for slot: %u",
+            __LINE__,
+            slot_nbr);
+      }
```
pyhys, line 2742: Should we `return ret; `at error? Also below.

src/device/pf_pdport.c
```diff
@@ -228,69 +299,117 @@ static void pf_pdport_remove_all_diag (pnet_t * net)
...
+void pf_pdport_remove_data_files (const char * file_directory)
+{
+   int port;
+   for (port = PNET_PORT_1; port <= PNET_MAX_PORT; port++)
+   {
+      pf_file_clear (file_directory, pf_pdport_get_filename (port));
+   }
+}
+
 void pf_pdport_ar_connected (pnet_t * net, const pf_ar_t * p_ar)
```
pyhys, line 352: Maybe `pf_pdport_ar_connect_ind()` 
Now it sounds like `bool pf_pdport_is_ar_connected()`
Just a suggestion

pyhys (commented):
Just two very quick fixes, then I'm happy.

src/device/pf_port.c
```diff
@@ -43,11 +43,25 @@ void pf_port_init(pnet_t * net)
...
+   pf_lldp_port_list_t * p_list)
 {
-   /* TODO: Implement support for multiple ports */
+   uint16_t port;
+   uint16_t index = 0;
+   uint8_t port_bit_mask = 0;
    memset (p_list, 0, sizeof (*p_list));
-   p_list->ports[0] = BIT (8 - 1);
+
+   CC_ASSERT (sizeof (*p_list) > (max_port / 8));
```
pyhys, line 56: To fix unittest:
`CC_ASSERT (sizeof (*p_list) >= (max_port / 8));`

src/device/pf_pdport.c
olbjo: Done

src/device/pf_port.c
olbjo: I moved the assert into the loop to avoid complex condition. 
Done

src/common/pf_alarm.c
olbjo: Updated so return code is -1 also if operation fails to get the module ids.
Done

src/device/pf_pdport.c
olbjo: Changed to pf_pdport_ar_connect_ind()
Done

pyhys (approved):
Great!

PR #274, opened 2020-12-11
Add option to configure size of port description

fredrikdanielmoller:
This increases the max size of port description
strings from 16 to 60 by default. The default
value of 60 is arbitrary, but hopefully large
enough to contain any received string.


pyhys (approved):
Great work, as always!

PR #275, opened 2020-12-13
Ensure consistency  for device location variable

fredrikdanielmoller:
The SNMP variable sysLocation should be the same as the I&M variable IM_Tag_Location.
This PR ensures that by letting SNMP server access IM_Tag_Location when sysLocation is requested.

The SNMP server executes from its own thread. A mutex is added for thread safety.

### DIFF
--- a/test/test_fspm.cpp
+++ b/test/test_fspm.cpp
@@ -20,6 +20,11 @@
 
 #include <gtest/gtest.h>
 
+class FspmTest : public PnetIntegrationTest
+{
+};
+
+
 class FspmUnitTest : public PnetUnitTest
 {
 };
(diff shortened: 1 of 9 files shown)


test/test_fspm.cpp
```diff
@@ -78,3 +83,44 @@ TEST_F (FspmUnitTest, FspmCheckValidateConfiguration)
...
+TEST_F (FspmTest, FspmGetSystemLocation)
+{
+   pf_snmp_system_location_t first;
+   pf_snmp_system_location_t second;
+
+   memset (&first, '1', sizeof (first));
+   memset (&second, '2', sizeof (second));
+
+   pf_fspm_get_system_location (net, &first);
+   EXPECT_LT (strlen (first.string), sizeof (first.string));
```
pyhys, line 96: Isn't this true for any null terminated string, regardless of content?
`EXPECT_LT (strlen (first.string), sizeof (first.string));`

pyhys (approved):
Great!
Just one minor question.

test/test_fspm.cpp
fredrikdanielmoller: Yes. The check tests that the returned string is null terminated.

pyhys: Ok, fine.

PR #276, opened 2020-12-14
DCP should use three different timers

pyhys:
The timers for clearing allowed MAC address, for flashing LED and
for sending a delayed response should be independent of eachother.

### DIFF
--- a/src/pf_types.h
+++ b/src/pf_types.h
@@ -927,7 +927,6 @@ typedef struct pf_alarm_queue
    uint16_t count;
 } pf_alarm_queue_t;
 
-
 #define PF_MAX_SESSION (2 * (PNET_MAX_AR) + 1) /* 2 per ar, and one spare. */
 
 /*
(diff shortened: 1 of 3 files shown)


src/pf_types.h
```diff
@@ -2691,8 +2691,9 @@ struct pnet
    pnet_ethaddr_t dcp_sam; /* Source address (MAC) of current DCP remote peer */
    bool dcp_delayed_response_waiting; /* A response to DCP IDENTIFY is waiting
                                          to be sent */
-   uint32_t dcp_timeout;
+   uint32_t dcp_led_timeout; /* Handle to the LED timeout instance */
    uint32_t dcp_sam_timeout; /* Handle to the SAM timeout instance */
+   uint32_t dcp_identresp_timeout; /* Handle to the SAM timeout instance */
```
olbjo, line 2696: Copy paste error in comment

pyhys: Done

olbjo (approved):

PR #289, opened 2021-01-04
Expose low-level diagnostics functionality

pyhys:
This is requested by users that use p-net for testing of
IO-controller applications.

Add configuration flag for default diagnosis mechanism
Also improve argument validation to diagnosis functions.

### DIFF
--- a/src/device/pf_diag.c
+++ b/src/device/pf_diag.c
@@ -47,11 +47,6 @@
 #include "pf_block_reader.h"
 #include "pf_block_writer.h"
 
-/** Which diagnosis mechanism to use when messaging the PLC.
- *  Normally PF_USI_QUALIFIED_CHANNEL_DIAGNOSIS would be used as it has
- *  more features, but it does not yet seem fully supported by Wireshark. */
-#define PF_DIAG_PREFERRED_DIAGNOSIS_USI PF_USI_EXTENDED_CHANNEL_DIAGNOSIS
-
 int pf_diag_init (void)
 {
    return 0;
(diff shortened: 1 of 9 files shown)


test/utils_for_testing.cpp
```diff
@@ -547,6 +547,9 @@ void PnetIntegrationTestBase::cfg_init()
 
    /* Storage */
    strcpy (pnet_default_cfg.file_directory, "/disk1");
+
+   /* Diagnosis */
+   pnet_default_cfg.use_qualified_diagnosis = true;
```
olbjo, line 552: I can't see that this config is set for the sample application. I think it should.

olbjo (approved):
I think the commit message could describe why the api was changed. It is not clear to me.

test/utils_for_testing.cpp
pyhys: Done

pyhys:
Improved the commit message.

src/device/pf_diag.c
```diff
@@ -720,22 +819,22 @@ int pf_diag_remove (
                    * the specifier PNET_DIAG_CH_SPEC_ALL_DISAPPEARS is used. */
                   /* Overwrite the item to remove with correct cr_properties for
                    * this diag message. */
+                  /* Why change USI? */
```
hefloryd, line 822: What does this comment refer to?

hefloryd (approved):

src/device/pf_diag.c
pyhys: It was a note to myself, now properly removed.

PR #294, opened 2021-01-08
snmp: Store sysName to file

fredrikdanielmoller:


src/common/pf_snmp.c
```diff
@@ -134,22 +134,80 @@ static void pf_snmp_encode_signal_delays (
...
    {
-      name->string[0] = '\0';
+      error = pnal_get_hostname (name->string);
+      if (error)
+      {
+         name->string[0] = '\0';
+      }
+      LOG_INFO (
+         PF_SNMP_LOG,
+         "SNMP(%d): Could not yet read sysName from nvm.\n",
```
pyhys, line 152: Maybe
`"SNMP(%d): Could not yet read sysName from nvm. Using hostname \"%s\".\n"`

src/common/pf_snmp.c
```diff
@@ -134,22 +134,80 @@ static void pf_snmp_encode_signal_delays (
...
 }
 
 int pf_snmp_set_system_name (pnet_t * net, const pf_snmp_system_name_t * name)
 {
-   /* TODO: Set new hostname permanently. Perhaps call pnal_set_ip_suite()? */
-   return -1;
+   const char * directory = net->fspm_cfg.file_directory;
+   pf_snmp_system_name_t temporary_buffer;
+   int res;
+
```
pyhys, line 167: `/* TODO: Possibly also set hostname via pnal_set_ip_suite() */`

pyhys (approved):
Great work as always!
Two minor comments (not required)

src/common/pf_snmp.c
fredrikdanielmoller: Done.

src/common/pf_snmp.c
fredrikdanielmoller: Setting hostname would probably not be correct as sysName (by convention) should be the fully qualified domain name, not the hostname.
Added comment instead.

PR #296, opened 2021-01-12
Improve timer handling in pf_dcp.c

pyhys:
Also better diagnostic printing of scheduler internals

This removes an invalid error message.


olbjo (approved):
👍

PR #300, opened 2021-01-13
Read Linux uptime for use in SNMP

pyhys:


fredrikdanielmoller (approved):

src/ports/linux/pnal.c
```diff
@@ -165,11 +166,14 @@ int pnal_load_file (
 
 uint32_t pnal_get_system_uptime_10ms (void)
 {
-   uint32_t uptime;
+   struct sysinfo systeminfo; /* Field .uptime contains uptime in seconds */
 
-   /* TODO: Get sysUptime from SNMP MIB-II */
-   uptime = 0;
-   return uptime;
+   if (sysinfo (&systeminfo) != 0)
```
hefloryd, line 171: This might not be the same value that is returned by sysUpTime. Is that OK? (sysUpTime is the time since the netsnmp-agent was restarted).

pyhys: The source https://oidref.com/1.3.6.1.2.1.1.3 says "The time (in hundredths of a second) since the network management portion of the system was last re-initialized."
I interpreted that as the Linux kernel, but maybe it should be net-snmp?

hefloryd: For the linux port that OID is implemented in system_mib.c and uses netsnmp_get_agent_uptime() which returns the time since the SNMP agent was restarted, which could be different from what sysinfo() returns.

pyhys: Used `sysinfo (&systeminfo)` for both PNAL and the Linux SNMP implementation.
Done.

PR #302, opened 2021-01-13
snmp: Fix error reporting on rt-kernel

fredrikdanielmoller:
* If requested variable from remote device does
  not exist, error NOSUCHINSTANCE is now reported
  back. Previously, a zero size value was sent.
* If requested variable exists, but its value is
  too large, error GENERROR is now reported back.
  Previously, a zero size value was sent.


pyhys (approved):
I have looked, but as I do not have the detailed SNMP knowledge I suggest that @hefloryd does a review.

PR #308, opened 2021-01-21
SNMP: Remove syscontact and sysname for application reset

pyhys:
For test case "Behavior scenario 10" when conformance class B is used.

### DIFF
--- a/src/device/pf_cmina.c
+++ b/src/device/pf_cmina.c
@@ -293,6 +293,13 @@ int pf_cmina_set_default_cfg (pnet_t * net, uint16_t reset_mode)
 
          /* Reset I&M data */
          ret = pf_fspm_clear_im_data (net);
+
+         /* According to section 8.4 "Behavior to ResetToFactory" in
+            "Test case specification: Behavior" the MIB data should be reset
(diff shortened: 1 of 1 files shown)


src/device/pf_cmina.c
```diff
@@ -293,6 +293,9 @@ int pf_cmina_set_default_cfg (pnet_t * net, uint16_t reset_mode)
 
          /* Reset I&M data */
          ret = pf_fspm_clear_im_data (net);
+
+         pf_file_clear (p_file_directory, PF_FILENAME_SYSCONTACT);
+         pf_file_clear (p_file_directory, PF_FILENAME_SYSNAME);
```
fredrikdanielmoller, line 298: The description for RESET_APPLICATION_DATA in PN-AL-Services ch. 6.3.11.3.2 does not state that sysContact or sysName should be reset. This is only stated in the description for RESET_COMMUNICATION_PARAMETER.

fredrikdanielmoller (approved):

src/device/pf_cmina.c
pyhys: Added comment in the code regarding additional requirement from test case specification.
Done.

PR #309, opened 2021-01-22
lldp: Only save received IPv4 address, not IPv6

fredrikdanielmoller:
Previously, if a received LLDP packet contained
multiple Management Address TLVs, only the last
one would be saved for later use. In the case
that the first TLV was the IPv4 address
and a second TLV contained an IPv6 address, the
IPv4 address would be discarded.
As Profinet uses Ipv4, this was bad.

Fixed by only saving the IPv4 address.

Unit tests were also added.

### DIFF
--- a/src/common/pf_lldp.c
+++ b/src/common/pf_lldp.c
@@ -538,6 +538,20 @@ void pf_lldp_reset_peer_timeout (
    }
 }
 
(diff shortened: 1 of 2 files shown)


pyhys (approved):
Excellent!

PR #312, opened 2021-01-25
Change ethernet interface name for rt-kernel sampleapp

pyhys:
Otherwise the test case Different Access Ways will complain.


fredrikdanielmoller (commented):
This doesn't seem correct. lwip will use a two character name assigned by the Ethernet driver, in this case "en", and then append a number, in this case 1. So "en1" should be correct.

pyhys:
The problem should be fixed in LWIP instead.

PR #314, opened 2021-01-26
Rename sampleapp variables

pyhys:
Subslots are numbered for example 0x8001, and we have around 5
positions in an array to store subslot data (for each slot).
Typically the variable 'subslot' is used for the subslot number.
Clarify the difference between the subslot number and the index into the
array.
    
Also improved NULL-check

### DIFF
--- a/sample_app/sampleapp_common.c
+++ b/sample_app/sampleapp_common.c
@@ -219,23 +219,29 @@ void app_print_network_details (
  * Get subslot application information.
  * @param p_appdata        InOut: Application state.
  * @param slot_nbr         In:    Slot number.
- * @param subslot_nbr      In:    Subslot number.
+ * @param subslot_nbr      In:    Subslot number. Range 0 - 0x9FFF.
  * @return Reference to application subslot,
- *         NULL is subslot is not found/plugged.
+ *         NULL if subslot is not found/plugged.
  */
 static app_subslot_t * app_subslot_get (
    app_data_t * p_appdata,
    uint16_t slot_nbr,
    uint16_t subslot_nbr)
 {
-   uint16_t subslot;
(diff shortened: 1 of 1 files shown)


sample_app/sampleapp_common.c
```diff
@@ -215,27 +215,43 @@ void app_print_network_details (
...
+ *         NULL if subslot is not found/plugged.
  */
 static app_subslot_t * app_subslot_get (
    app_data_t * p_appdata,
    uint16_t slot_nbr,
    uint16_t subslot_nbr)
 {
-   uint16_t subslot;
-   for (subslot = 0; subslot < PNET_MAX_SUBSLOTS; subslot++)
+   uint16_t subslot_array_index; /* Range 0 - (PNET_MAX_SUBSLOTS - 1) */
```
olbjo, line 238: I see that the name "sub_ix" is used at least at one time in the stack. Maybe such a shorter name is to prefer? Or at least shorted to "subslot_index".

Also I think that the range is quite clear given the conditions on the for loop. I do not think the comment adds much.

olbjo (approved):
As you see by my comments I don't agree on the renaming.
Go ahead and merge if you want to. I do not see any errors. 😄

sample_app/sampleapp_common.c
```diff
@@ -215,27 +215,43 @@ void app_print_network_details (
 
 /************************** Subslot data *************************************/
 
+/* TODO: We do not unplug modules from slots at release or abort.
```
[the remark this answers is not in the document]
pyhys: Removed.
Done.

sample_app/sampleapp_common.c
pyhys: Good point. 
Renamed to `subslot_ix`
Done

olbjo:
Great!

PR #320, opened 2021-01-28
rt-kernel: Update MIB2 counters

fredrikdanielmoller:
Let Ethernet driver update counters in SNMP MIB2
table for each sent and for each received Ethernet
frame.

This change was taken from the rt-kernel repository (version 2020.1).

### DIFF
--- a/src/ports/rt-kernel/dwmac1000.c
+++ b/src/ports/rt-kernel/dwmac1000.c
@@ -23,7 +23,7 @@
 #include <lwip/stats.h>
 #include <lwip/tcpip.h>
 #include <lwip/opt.h>
-
+#include <lwip/snmp.h>
 #include <netif/etharp.h>
 
 #include <string.h>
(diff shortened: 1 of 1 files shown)


hefloryd (commented):
Could we add p-net changes to the patch instead of having a copy of dwmac1000.c in this repo?

fredrikdanielmoller:
> Could we add p-net changes to the patch instead of having a copy of dwmac1000.c in this repo?

Yes. I suggest we do this in a future PR.
It would also be nice if the p-net specific callback input_rx_hook() could be replaced with LWIP_HOOK_UNKNOWN_ETH_PROTOCOL.

hefloryd:
> > Could we add p-net changes to the patch instead of having a copy of dwmac1000.c in this repo?
> 
> Yes. I suggest we do this in a future PR.
> It would also be nice if the p-net specific callback input_rx_hook() could be replaced with LWIP_HOOK_UNKNOWN_ETH_PROTOCOL.

I think we looked into this previously. It should work but IIRC requires one or two extra context switches, so might affect performance. @nakarlsson knows more?

PR #322, opened 2021-02-01
snmp: Don't return cached values on rt-kernel

fredrikdanielmoller:
Previously, SNMP Get/GetNext requests for MIB-II
system variables were handled by lwip by either
returning values read at startup or the values
later set in SNMP Set requests. This did not allow
for the case that system variables were changes by
other means than SNMP.

To solve this, the lwip handling of MIB-II system
variables is replaced with a custom version.

Stack size for SNMP thread was also increased as
pf_snmp_get_*() functions are now called by the
SNMP thread instead of main thread, plus that use
of some static variables were eliminated.

### DIFF
--- a/cmake/rt-kernel.cmake
+++ b/cmake/rt-kernel.cmake
@@ -24,6 +24,7 @@ target_sources(profinet
   src/ports/rt-kernel/pnal_eth.c
   src/ports/rt-kernel/pnal_udp.c
   src/ports/rt-kernel/pnal_snmp.c
+  src/ports/rt-kernel/mib2_system.c
   src/ports/rt-kernel/lldp-mib.c
   src/ports/rt-kernel/lldp-ext-pno-mib.c
   src/ports/rt-kernel/lldp-ext-dot3-mib.c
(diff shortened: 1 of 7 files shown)


pyhys (approved):
Great work, as always!

src/ports/rt-kernel/snmp_mib2.c
```diff
@@ -0,0 +1,137 @@
+/*********************************************************************
```
hefloryd, line 1: I don't think we can add our own header to this file. BTW, we will update the cmake toolchain so that linker should be able to resolve circular dependencies (by wrapping all libs in -Wl,--start-group/-Wl,--end-group), if that would help.

fredrikdanielmoller: Ok, I'll remove the header.

The mentionened cmake change could make it possible to remove this file.

fredrikdanielmoller:
Updated the patch file. lwip now calls callback functions when system variables are read (or written) in MIB-II.

src/ports/rt-kernel/snmp_mib2.c
fredrikdanielmoller: Entire file removed. Similar header was also removed in another file.

PR #324, opened 2021-02-04
Git should not do line end normalization on bash scripts

pyhys:


olbjo (approved):
I tested by cloning a fresh repo in windows and switching branches and checked the line endings.

PR #327, opened 2021-02-05
Use syncronized file saving on Linux 

pyhys:
This is required for file changes to appear on disk, even if the
power is removed soon after writing. Tested with ART tester test
case "Topology discovery check".

Bug reported by Giuseppe.

### DIFF
--- a/src/ports/linux/pnal.c
+++ b/src/ports/linux/pnal.c
@@ -35,6 +35,7 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <limits.h>
 #include <signal.h>
 #include <stdarg.h>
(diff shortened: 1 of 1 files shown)


src/ports/linux/pnal.c
```diff
@@ -97,6 +97,7 @@ int pnal_save_file (
 
    /* Close file */
    fclose (outputfile);
+   sync();
```
olbjo, line 100: Maybe fsync() is better. Syncing/flushing whole file system should probably be avoided.
I think we should rerun test with fsync to check if that works and also compare the timing of  the two

olbjo (changes requested):

src/ports/linux/pnal.c
pyhys: Ok, I will try.

pyhys: The `fsync()` approach did unfortunately not work, but I removed the `sync()` from `pnal_clear_file()`.

olbjo: OK

olbjo (approved):
👍

src/ports/linux/pnal.c
hefloryd: It might be worth investigating O_SYNC | O_DIRECT.

pyhys: `fflush()` and `fsync()` together did the trick.

pyhys:
Changed to use non-buffered kernel API.

pyhys:
Remove redundant `fsync` when using `O_SYNC`

Issue #316, opened 2021-01-26
Minimise configuration structure


fredrikdanielmoller:
The configuration structure contains fields which could be calculated in pnet_init(). Remove these fields from pnet_cfg_t as to simplify stack configuration for the user.

Fields which could removed from configuration:
* eth_addr in pnet_netif_t. Initialise by calling pnal_get_macaddress().
* port_id in pnet_port_cfg_t. Initialise as "port-001", "port-002" etc.
* ip_cfg in pnet_if_cfg_t. Initialise by calling pnal_get_ip_address()/subnet()/gateway().
* Parts of im_x_data.

fredrikdanielmoller:
It would also be better if the configuration structure was only used as a user-visible structure optimised for user ease-of-use. Currently, it serves a double purpose as it is also stored in the `pnet_t` instance structure where its members are read and written.

PR #333, opened 2021-02-10
Simplify configuration

olbjo:
Implements part of enhancements suggested in #316.

Main change is that ethernet address and port name are not configuration parameters set by sample application.
To review the the update check out changes in pnet_api.h and pf_types.h rest of the updates follows from these updates.

A second commit renames PNET_MAX_PORT to PNET_NUMBER_OF_PHYSICAL_PORTS

Rebased on #332 to avoid future merge conflict.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,12 +60,12 @@ set(PNET_MAX_CR                 2
 set(PNET_MAX_SLOTS            5
   CACHE STRING "Per API. Should be > 1 to allow at least one I/O module")
 set(PNET_MAX_SUBSLOTS         3
-  CACHE STRING "Per slot (DAP requires 2 + PNET_MAX_PORT)")
+  CACHE STRING "Per slot (DAP requires 2 + PNET_NUMBER_OF_PHYSICAL_PORTS)")
 set(PNET_MAX_CHANNELS           1
   CACHE STRING "Per sub-slot. Used for diagnosis")
 set(PNET_MAX_DFP_IOCR           2
   CACHE STRING "Allowed values are 0 (zero) or 2")
-set(PNET_MAX_PORT               1
+set(PNET_NUMBER_OF_PHYSICAL_PORTS          1
   CACHE STRING "Number of physical ports")
 set(PNET_MAX_LOG_BOOK_ENTRIES   16
   CACHE STRING "")
--- a/include/pnet_api.h
+++ b/include/pnet_api.h
@@ -1171,23 +1171,11 @@ typedef struct pnet_ethaddr
    (PNET_STATION_NAME_MAX_SIZE + PNET_PORT_NAME_MAX_SIZE)
 
 /**
- * Network interface
- */
-typedef struct pnet_netif
-{
-   char if_name[PNET_INTERFACE_NAME_MAX_SIZE]; /**< Terminated string */
-   pnet_ethaddr_t eth_addr;                    /**< Interface MAC address */
-} pnet_netif_t;
-
-/**
  * Physical Port Configuration
  */
 typedef struct pnet_port_cfg
 {
-   pnet_netif_t phy_port;
-   char port_name[PNET_PORT_NAME_MAX_SIZE]; /**< Terminated string */
-   uint16_t rtclass_2_status;
-   uint16_t rtclass_3_status;
+   const char * netif_name;
 } pnet_port_cfg_t;
 
 /**
@@ -1213,10 +1201,10 @@ typedef struct pnet_ip_cfg
  */
 typedef struct pnet_if_cfg
 {
-   pnet_netif_t main_port; /**< Main (DAP) network interface. */
+   const char * main_netif_name; /**< Main (DAP) network interface. */
    pnet_ip_cfg_t ip_cfg;   /**< IP Settings for main network interface */
 
-   pnet_port_cfg_t ports[PNET_MAX_PORT]; /**< Physical ports (DAP ports) */
+   pnet_port_cfg_t physical_ports[PNET_MAX_PORT];
 } pnet_if_cfg_t;
 
 /**
--- a/include/pnet_api.h
+++ b/include/pnet_api.h
@@ -1204,7 +1204,7 @@ typedef struct pnet_if_cfg
    const char * main_netif_name; /**< Main (DAP) network interface. */
    pnet_ip_cfg_t ip_cfg;   /**< IP Settings for main network interface */
 
-   pnet_port_cfg_t physical_ports[PNET_MAX_PORT];
+   pnet_port_cfg_t physical_ports[PNET_NUMBER_OF_PHYSICAL_PORTS];
 } pnet_if_cfg_t;
 
 /**
--- a/src/common/pf_eth.c
+++ b/src/common/pf_eth.c
@@ -33,17 +33,6 @@
 #include <string.h>
 #include "pf_includes.h"
 
-/**
- * Init network interface.
- * Set name, read mac address, register receive callback.
- *
- * @param net              InOut: The p-net stack instance
- * @param netif_name       In:    Network interface name
- * @param eth_receive_type In:    Ethernet frame types that shall be received
- *                                by the network interface / port.
- * @param netif            Out:   Initialized network interface.
- * @return
- */
 static int pf_eth_init_netif (
    pnet_t * net,
    const char * netif_name,
@@ -83,13 +72,13 @@ int pf_eth_init (pnet_t * net, const pnet_cfg_t * p_cfg)
    pf_port_iterator_t port_iterator;
    pf_port_t * p_port_data;
    pnal_ethertype_t main_port_receive_type;
-#if PNET_MAX_PORT > 1
+#if PNET_NUMBER_OF_PHYSICAL_PORTS > 1
    const pnet_port_cfg_t * p_port_cfg;
 #endif
 
    memset (net->eth_id_map, 0, sizeof (net->eth_id_map));
 
-#if (PNET_MAX_PORT == 1)
+#if (PNET_NUMBER_OF_PHYSICAL_PORTS == 1)
    main_port_receive_type = PNAL_ETHTYPE_ALL;
 #else
    main_port_receive_type = PNAL_ETHTYPE_PROFINET;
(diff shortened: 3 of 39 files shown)


src/common/pf_eth.c
```diff
@@ -69,27 +109,23 @@ int pf_eth_init (pnet_t * net, const pnet_cfg_t * p_cfg)
...
+            &p_port_data->netif) != 0)
       {
-         LOG_ERROR (
-            PNET_LOG,
-            "Failed to init \"%s\"\n",
-            p_port_cfg->phy_port.if_name);
          return -1;
       }
+#else
+      /* In single port configuration managed port is also physical port 1 */
```
pyhys, line 115: Maybe
`/* In single port configuration is managed port also physical port 1 */`

src/device/pf_port.c
```diff
@@ -39,6 +39,11 @@ void pf_port_init (pnet_t * net)
    {
       p_port_data = pf_port_get_state (net, port);
       p_port_data->port_num = port;
+      snprintf (
+         p_port_data->port_name,
+         sizeof (p_port_data->port_name),
+         "port-%03u",
+         (uint8_t)port);
```
pyhys, line 46: Maybe a comment why we need to cast?

CMakeLists.txt
```diff
@@ -60,12 +60,12 @@ set(PNET_MAX_CR                 2
 set(PNET_MAX_SLOTS            5
   CACHE STRING "Per API. Should be > 1 to allow at least one I/O module")
 set(PNET_MAX_SUBSLOTS         3
-  CACHE STRING "Per slot (DAP requires 2 + PNET_MAX_PORT)")
+  CACHE STRING "Per slot (DAP requires 2 + PNET_MAX_PHY_PORTS)")
```
pyhys, line 63: I prefer not to have `_MAX_` in the name, simply because if we have one port then it is not valid to run with `PNET_MAX_PHY_PORTS = 2`
Maybe
`PNET_PHY_PORTS` or `PNET_NUM_PHY_PORTS` or `PNET_NUMBER_OF_PORTS`

pyhys (commented):
Great work, this really simplifies the usage of the library.
Just a few trivial comments.

pyhys:
Part of work with #316

include/pnet_api.h
```diff
@@ -1171,23 +1171,19 @@ typedef struct pnet_ethaddr
...
 /**
  * Physical Port Configuration
  */
 typedef struct pnet_port_cfg
 {
-   pnet_netif_t phy_port;
-   char port_name[PNET_PORT_NAME_MAX_SIZE]; /**< Terminated string */
-   uint16_t rtclass_2_status;
-   uint16_t rtclass_3_status;
+   pnet_netif_name_t port_netif;
```
fredrikdanielmoller, line 1186:
```suggestion
   const char * network_interface_name;
```
(Assuming that pnet_cfg_t is no longer stored in pnet_t.)

include/pnet_api.h
```diff
@@ -1213,10 +1209,11 @@ typedef struct pnet_ip_cfg
  */
 typedef struct pnet_if_cfg
 {
-   pnet_netif_t main_port; /**< Main (DAP) network interface. */
+   pnet_netif_name_t main_netif; /**< Main (DAP) network interface. */
    pnet_ip_cfg_t ip_cfg;   /**< IP Settings for main network interface */
 
-   pnet_port_cfg_t ports[PNET_MAX_PORT]; /**< Physical ports (DAP ports) */
+   pnet_port_cfg_t phy_ports[PNET_MAX_PHY_PORTS]; /**< Physical ports (DAP
```
fredrikdanielmoller, line 1215:
```suggestion
   pnet_port_cfg_t physical_ports[PNET_MAX_PHY_PORTS]; /**< Physical ports (DAP
```

include/pnet_api.h
```diff
@@ -1213,10 +1209,11 @@ typedef struct pnet_ip_cfg
  */
 typedef struct pnet_if_cfg
 {
-   pnet_netif_t main_port; /**< Main (DAP) network interface. */
+   pnet_netif_name_t main_netif; /**< Main (DAP) network interface. */
```
fredrikdanielmoller, line 1212:
```suggestion
   const char * network_interface_name; /**< Main (DAP) network interface. */
```
(Assuming that pnet_cfg_t is no longer stored in pnet_t.)

fredrikdanielmoller (commented):
For strings in configuration, I think it would be more user-friendly to let them be pointers instead of arrays (i.e. `const char*` instead of `char[N`]).

pyhys (commented):
Add this to src/device/pf_port.c:
```
#if PNET_MAX_SUBSLOTS < (2 + PNET_NUMBER_OF_PHYSICAL_PORTS)
#error "DAP requires 2 + PNET_NUMBER_OF_PHYSICAL_PORTS subslots. Increase PNET_MAX_SUBSLOTS."
#endif
```

src/common/pf_eth.c
olbjo: Rephrased, done

include/pnet_api.h
olbjo: Remove pnet_cfg_t from pnet_t is out of the scope for this task. I will use name "netif_name".

include/pnet_api.h
olbjo: Done

include/pnet_api.h
olbjo: will use "netif_name"

include/pnet_api.h
fredrikdanielmoller: Beware that the copy of pnet_cfg_t in pnet_t will be a shallow copy, unless the strings are also copied.

src/device/pf_port.c
olbjo: Done

olbjo:
Comments handled, rebased on master. @pyhys I think you need to re-review updaes

pyhys (approved):
Great!

CMakeLists.txt
```diff
@@ -60,12 +60,12 @@ set(PNET_MAX_CR                 2
...
 set(PNET_MAX_CHANNELS           1
   CACHE STRING "Per sub-slot. Used for diagnosis")
 set(PNET_MAX_DFP_IOCR           2
   CACHE STRING "Allowed values are 0 (zero) or 2")
-set(PNET_MAX_PORT               1
+set(PNET_MAX_PHY_PORTS          1
```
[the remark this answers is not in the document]
pyhys: This has been done

include/pnet_api.h
```diff
@@ -1213,10 +1201,10 @@ typedef struct pnet_ip_cfg
  */
 typedef struct pnet_if_cfg
 {
-   pnet_netif_t main_port; /**< Main (DAP) network interface. */
+   const char * main_netif_name; /**< Main (DAP) network interface. */
```
fredrikdanielmoller, line 1204: In `pf_fspm_init()`, a copy of the configuration will be stored. When the string was allocated as an array in `pnet_cfg_t`, this would automatically copy the string. As the sting is now a pointer, no such copy will be made. This may be a problem if user allocated the string in non-static memory, as the content of the string could then change after the call to `pnet_init()`.

This could be solved by requiring that user allocates the string in static storage, or by making a copy of the string in `pf_fspm_init()`. I would recommend the second option, as it is more user-friendly.

src/pf_types.h
```diff
@@ -2728,14 +2728,20 @@ typedef struct pf_lldp_port
    bool is_peer_info_received;
 } pf_lldp_port_t;
 
+typedef struct pf_netif
+{
+   char name[PNET_INTERFACE_NAME_MAX_SIZE]; /**< Terminated string */
+   pnet_ethaddr_t mac;
+   pnal_eth_handle_t * eth_handle;
```
fredrikdanielmoller, line 2735:
```suggestion
   pnal_eth_handle_t * handle;
```
Note that at point-of-use, this will read `netif.handle`.

src/pf_types.h
```diff
@@ -2728,14 +2728,20 @@ typedef struct pf_lldp_port
    bool is_peer_info_received;
 } pf_lldp_port_t;
 
+typedef struct pf_netif
```
fredrikdanielmoller, line 2731:
```suggestion
/** Network interface */
typedef struct pf_netif
```

src/pf_types.h
```diff
@@ -2728,14 +2728,20 @@ typedef struct pf_lldp_port
...
+{
+   char name[PNET_INTERFACE_NAME_MAX_SIZE]; /**< Terminated string */
+   pnet_ethaddr_t mac;
```
fredrikdanielmoller, line 2734:
```suggestion
   pnet_ethaddr_t mac_address;
```

src/pf_types.h
```diff
@@ -2811,12 +2816,13 @@ struct pnet
     */
    struct
    {
+      pf_netif_t main_port;
```
fredrikdanielmoller, line 2819:
```suggestion
      pf_netif_t main_port; /* Management port */
```

fredrikdanielmoller (changes requested):
Reviewed pnet_api.h and pf_types.h

src/pf_types.h
olbjo: Done

src/pf_types.h
olbjo: Done

src/pf_types.h
olbjo: Done

src/pf_types.h
olbjo: Done

include/pnet_api.h
olbjo: Netif names part of p_cfg will be copied into net->pf_interface data structure during initialization. I removed all other use of netif names in p_cfg. For example pf_port_get_state() is used instead of pf_port_get_config().
Done

fredrikdanielmoller (approved):

CMakeLists.txt
pyhys: Name is now changed

olbjo:
Removed occurrence of PNET_MAX_PORT that emerged due to rebase.

PR #338, opened 2021-02-11
Rename "interface" struct

olbjo:
To avoid problems with Microsoft compilers

### DIFF
--- a/src/pf_types.h
+++ b/src/pf_types.h
@@ -2817,7 +2817,7 @@ struct pnet
          pf_lldp_name_of_device_mode_t mode;
       } name_of_device_mode;
       pf_port_t port[PNET_MAX_PORT];
-   } interface;
+   } pf_interface;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


pyhys (approved):
Great

fredrikdanielmoller:
Where does "dap" come from? "device access point" does not sound familiar. I searched for it in PN-AL-Protocol without success.

olbjo:
"DAP is the access point for communication with the Ethernet interface and the processing program. "

The absence of a definition or description of DAP in the PN-AL Protocol I can not understand. Most whitepapers on Profinet refers to it. The first data slot is the DAP slot and its sub-slot provides the interface for control of device, interface and port settings. All of which I think fits nicely into this data structure.

src/pf_types.h
```diff
@@ -2822,7 +2822,7 @@ struct pnet
          pf_lldp_name_of_device_mode_t mode;
       } name_of_device_mode;
       pf_port_t port[PNET_MAX_PORT];
-   } interface;
+   } dap;
```
hefloryd, line 2825: Why not prefix with pf_ or similar? Could even be pf_interface if that is more clear than pf_dap?

olbjo: Changed to "pf_interface".

Done

fredrikdanielmoller (approved):

PR #340, opened 2021-02-12
rt-kernel: Fix ifSpeed in SNMP MIB-II ifTable

fredrikdanielmoller:
ifSpeed would always be reported as 0 bps. Fixed by setting it in Ethernet driver's probe function.


pyhys (approved):

pyhys:
I think @hefloryd needs to look  on the details.

PR #348, opened 2021-02-17
Many alias names

olbjo:
Handle multiple alias names.

PR #349, opened 2021-02-17
Soft mac filtering

olbjo:
In promiscuous mode, ignore unicast packets that are not for this device.
Based on top of #333 and #348

### DIFF
--- a/src/common/pf_eth.c
+++ b/src/common/pf_eth.c
@@ -152,6 +152,18 @@ int pf_eth_recv (pnal_eth_handle_t * eth_handle, void * arg, pnal_buf_t * p_buf)
    uint16_t ix = 0;
    pnet_t * net = (pnet_t *)arg;
 
+#ifndef UNIT_TEST
+   /* Reject unicast packets with destination not matching
+    * mac address of management interface.
+    */
+   if (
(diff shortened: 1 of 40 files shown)


pyhys (approved):
Tested with ART tester, and it works fine

src/common/pf_eth.c
```diff
@@ -122,6 +149,18 @@ int pf_eth_recv (pnal_eth_handle_t * eth_handle, void * arg, pnal_buf_t * p_buf)
    uint16_t ix = 0;
    pnet_t * net = (pnet_t *)arg;
 
+#ifndef UNIT_TEST
+   /* Reject unicast packets with destination not matching
```
fredrikdanielmoller, line 153: Wouldn't it be better if unicast filtering was done at a lower level (ie pnal)? Documentation could be added to pnal_eth_init() to require that incoming frames with unknown unicast destination MAC address should be dropped.

fredrikdanielmoller (changes requested):

src/common/pf_eth.c
olbjo: I agree, we will start with only stop using promiscuous mode and retest to verify that this solves the original problem.

PR #352, opened 2021-02-18
Do not use promiscuous mode

olbjo:


pyhys (approved):
I did run all ART-tester test cases with a branch having this change.

PR #353, opened 2021-02-18
Update OSAL and gtest

hefloryd:
This PR updates gtest to fix cmake deprecation warnings, and to fix a runtime warning on embedded targets. Unfortunately gtest no longer builds on GCC version 4 or older due to incomplete c++11 support. Use of incompatible GCC is detected at configuration time and the user can disable test by setting BUILD_TESTING=OFF.

It also updates OSAL and reduces the number of global cmake settings being forced, to simplify including this project in a superproject.


pyhys (approved):
Sweet!

PR #355, opened 2021-02-19
Fix for test case SM_Legacy

olbjo:
To enable the test for legacy startup mode,
in the gsdml file, set the StartupMode attribute to "Legacy;Advanced"


pyhys:
See #381

PR #359, opened 2021-02-21
rt-kernel: Use patched Ethernet driver

fredrikdanielmoller:
The patch file was updated to include p-net specific
changes to the xmc4 Ethernet driver, dwmac1000.c.
The dwmac1000.c file in port directory was removed.

The function eth_ioctl() was replaced with driver
specific ioctl callback.


src/ports/rt-kernel/pnal.c
```diff
@@ -169,26 +163,26 @@ static pnal_eth_mau_t calculate_mau_type (uint8_t link_state)
 /**
  * Calculate advertised capabilites
  *
- * @param anar       In:    Contents of the ANAR register read from PHY.
+ * @param rtkernel_capabilities In:    Contents of the ANAR register read from PHY.
```
hefloryd, line 166: advertised or advertised_capabilities? The capabilities are not related to rt-kernel?

fredrikdanielmoller: Changed to `capabilities` as to match defines. Also updated parameter comment as to clarify.

PR #361, opened 2021-02-22
Remove diagnosis when peer returns

pyhys:


src/device/pf_pdport.c
```diff
@@ -752,6 +787,7 @@ static void pf_pdport_run_peer_check (pnet_t * net, int loc_port_num)
 
    if (p_port_data->pdport.check.active == true)
    {
+      pf_pdport_remove_missing_peer_diagnosis (net, loc_port_num);
```
olbjo, line 790: Maybe better name is "pf_pdport_check_missing_peer_diagnosis" ?

olbjo: Or even better:
we remove "missing peer" in logs/naming/comments, 
The alarm is named "no peer detected" and we should use this instead.

So, new  suggested name is : "pf_pdport_check_no_peer_detected_diagnosis "

also replace use of "missing peer" with "no peer detected"

olbjo (changes requested):
Great!
Name changes suggested. Lets discuss.

src/device/pf_pdport.c
pyhys: Done

olbjo (approved):
Great!

PR #371, opened 2021-02-25
Sort diagnoses in USI order at readout

pyhys:
Insert a proper block header before each group of diagnoses.
Note that all diagnoses in a block should have the same
API, slot, subslot and USI value.

### DIFF
--- a/src/device/pf_block_writer.c
+++ b/src/device/pf_block_writer.c
@@ -2535,6 +2535,8 @@ void pf_put_log_book_data (
  *  - Extended channel diagnosis (standard format)
  *  - Qualified channel diagnosis (standard format)
  *
+ * Used both for creating an alarm frame and a diagnostics read answer frame.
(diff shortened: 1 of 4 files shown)


src/device/pf_block_writer.c
```diff
@@ -2762,33 +2789,150 @@ static void pf_put_diag_list (
...
+         *p_pos = block_pos;
+      }
+   }
+}
+
+/**
+ * @internal
+ * Find next diagnosis USI value (sorted) for a subslot
+ *
+ * Finds the smallest USI value that is larger than \a low_usi_limit
```
olbjo, line 2832: greater?

olbjo (approved):
Looks great! 
A minor comment on the comments

pyhys:
Moved `get_next_diagnosis_usi()` to `pf_cmdev.c`

PR #372, opened 2021-02-26
Do not overwrite errorcodes

pyhys:
When writing I&M data, the errorcodes set in one function were
later overwritten in another function. Fix that.

Only write I&M to file after valid requests.

Clean up error-code related defines, and some error-code related
comments.

Found with test case IP_UDP_RPC_I&M_EPM

### DIFF
--- a/src/device/pf_cmrpc.c
+++ b/src/device/pf_cmrpc.c
@@ -787,16 +787,77 @@ void pf_set_error (
 {
    LOG_INFO (
       PNET_LOG,
-      "CMRPC(%d) ERROR: %u, %u, %u, %u\n",
+      "CMRPC(%d) Setting error: 0x%02X 0x%02X 0x%02X 0x%02X\n",
       __LINE__,
       (unsigned)code,
       (unsigned)decode,
       (unsigned)code_1,
       (unsigned)code_2);
+
+   if (
+      p_stat->pnio_status.error_code != PNET_ERROR_CODE_NOERROR ||
+      p_stat->pnio_status.error_decode != PNET_ERROR_DECODE_NOERROR)
+   {
+      LOG_DEBUG (
+         PNET_LOG,
+         "CMRPC(%d) Did overwrite previous error: 0x%02X 0x%02X 0x%02X "
+         "0x%02X\n",
+         __LINE__,
+         (unsigned)p_stat->pnio_status.error_code,
+         (unsigned)p_stat->pnio_status.error_decode,
+         (unsigned)p_stat->pnio_status.error_code_1,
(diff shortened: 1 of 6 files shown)


src/device/pf_cmrpc.c
```diff
@@ -787,7 +787,7 @@ void pf_set_error (
 {
    LOG_INFO (
       PNET_LOG,
-      "CMRPC(%d) ERROR: %u, %u, %u, %u\n",
+      "CMRPC(%d) Setting error: 0x%02X 0x%02X 0x%02X 0x%02X\n",
```
olbjo, line 790: Maybe it is a good idea to also log if another error is already set and the currently set error code?

src/device/pf_cmrpc.c
```diff
@@ -799,6 +799,33 @@ void pf_set_error (
...
+         __LINE__,
+         (unsigned)code,
+         (unsigned)decode,
+         (unsigned)code_1,
+         (unsigned)code_2);
+      p_stat->pnio_status.error_code = code;
+      p_stat->pnio_status.error_decode = decode;
+      p_stat->pnio_status.error_code_1 = code_1;
+      p_stat->pnio_status.error_code_2 = code_2;
+   }
```
olbjo, line 826: Add log of current error as well as ignored error code?

src/device/pf_cmrpc.h
```diff
@@ -80,6 +80,23 @@ void pf_set_error (
...
+ * Insert detailed error information into the result structure (of a session).
+ * Does not overwrite existing error information.
+ *
+ * @param p_stat           Out:  The result structure.
+ * @param code             In:   The error_code.
+ * @param decode           In:   The error_decode.
+ * @param code_1           In:   The error_code_1.
+ * @param code_2           In:   The error_code_2.
+ */
+void pf_set_error_if_not_already_set (
```
olbjo, line 93: Is there a way to know when to use this function or the default pf_set_error()? If so a comment describing such rules is very helpful. Current solution is a bit of a workaround.

olbjo (approved):
👍 
I see no errors but some comments on the logging

src/device/pf_cmrpc.c
pyhys: Done

src/device/pf_cmrpc.c
pyhys: Done

src/device/pf_cmrpc.h
pyhys: Done

PR #374, opened 2021-03-01
Set SO_REUSEADDR for UDP sockets to enable fast restarts

### LINKED ISSUE #315: Correct shutdown sequence
I am using the library for my slave device. I am experiencing troubles when I run two consecutive (non-parallel) runs: while the first goes without problems, the second one never receives the State Indication change and causes a "00006001 RPC_Request_Blocked" on the Codesys Control for Linux instance. I am issuing the pnet_ar_abort command to then wait for the Abort State Indication, after that I just stop calling the pnet_handle_periodic(). What I don't really understand is if this is the right way to handle the slave disconnection, leaving things clean for the next instance to be executed. Thanks.

pyhys:
Closes #315

### DIFF
--- a/src/ports/linux/pnal_udp.c
+++ b/src/ports/linux/pnal_udp.c
@@ -21,37 +21,39 @@
 
 int pnal_udp_open (pnal_ipaddr_t addr, pnal_ipport_t port)
 {
-   int ret = -1;
    struct sockaddr_in local;
    int id;
+   const int enable = 1;
 
    id = socket (PF_INET, SOCK_DGRAM, IPPROTO_UDP);
-
-   if (id < 0)
+   if (id == -1)
    {
-      ret = id;
-   }
-   else
-   {
-      /* set IP and port number */
-      local = (struct sockaddr_in){
-         .sin_family = AF_INET,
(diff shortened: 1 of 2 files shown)


src/ports/linux/pnal_udp.c
```diff
@@ -40,14 +37,21 @@ int pnal_udp_open (pnal_ipaddr_t addr, pnal_ipport_t port)
          .sin_port = htons (port),
          .sin_zero = {0},
       };
-      ret = bind (id, (struct sockaddr *)&local, sizeof (local));
-      if (ret != 0)
+
+      if (setsockopt(id, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(enable)) < 0)
```
olbjo, line 41: != 0

Reading the refman setsockopt() should return 0 on success. 
Any reason for not using same condition as for bind()?

olbjo (approved):
👍 
A comment on setsockopt() return code.

src/ports/linux/pnal_udp.c
pyhys: Done

PR #375, opened 2021-03-02
Detect link down

pyhys:


src/device/pf_pdport.c
```diff
@@ -787,48 +809,120 @@ static void pf_pdport_run_peer_check (pnet_t * net, int loc_port_num)
...
+         __LINE__,
+         loc_port_num);
+      return;
+   }
+
+   if (p_port_data->netif.previous_is_link_up == false && status.running)
+   {
+      pf_pdport_handle_link_up (net, loc_port_num);
+   }
+   else if (p_port_data->netif.previous_is_link_up && status.running == false)
```
olbjo, line 897: !status.running   ?
since we line 893  is not status.running == true

src/device/pf_pdport.c
```diff
@@ -787,48 +809,120 @@ static void pf_pdport_run_peer_check (pnet_t * net, int loc_port_num)
...
 }
 
 void pf_pdport_periodic (pnet_t * net)
 {
    int port;
    pf_port_iterator_t port_iterator;
    pf_port_t * p_port_data = NULL;
+   static int counter = 0;
+
+   // TODO Move somewhere else
```
olbjo, line 912: idea:
Move monitor functionality into separate function which is registered to scheduler, run every 100ms. 

Maybe we should measure the time it takes to make the system call. If needed, only poll status of one network interface at a time. The the port_iterator variable can be stored in port data and only one port is checked every iteration.

On the other hand, if system call is fast we move pf_pdport_monitor_link (net, port); into 
existing iterator loop which would be simplest solution.

olbjo (commented):
Great!
See notes on the todo.

pyhys (commented):
Implemented a scheduled function that scan the Ethernet links

src/device/pf_pdport.c
pyhys: Implemented a scheduled function that scan the Ethernet links

olbjo (approved):
👍 
When I see the code I think that using separate scheduler timeout for each port would have saved you a some work..

PR #378, opened 2021-03-09
Trigger PNET_EVENT_PRMEND after DControl response have been sent

pyhys:


src/device/pf_cmdev.c
```diff
@@ -4869,7 +4869,8 @@ int pf_cmdev_rm_dcontrol_ind (
    pnet_t * net,
    pf_ar_t * p_ar,
    pf_control_block_t * p_control_io,
-   pnet_result_t * p_dcontrol_result)
+   pnet_result_t * p_dcontrol_result,
+   bool * p_set_state_paramend)
```
olbjo, line 4873: paramend too similar to PRMEND? use param_end or parameter_end or prmend

sample_app/sampleapp_common.c
```diff
@@ -825,7 +825,15 @@ static int app_state_ind (
       (void)pnet_set_provider_state (net, true);
 
       /* Send application ready  */
-      os_event_set (p_appdata->main_events, APP_EVENT_READY_FOR_DATA);
+      if (p_appdata->arguments.verbosity > 0)
```
olbjo, line 828: Maybe the old event shall be kept? 
An application that needs additional time to prepare before sending the application ready can then use the existing event to implement this.  This is a comment. I am not sure what is the best solution

olbjo (approved):

sample_app/sampleapp_common.c
pyhys: Ok. 
Done.

src/device/pf_cmdev.c
pyhys: Now `p_set_state_prmend`
Done

PR #379, opened 2021-03-09
Use configurable number of ports

pyhys:
This allows the same precompiled library to be used for different number of ports

### DIFF
--- a/test/test_port.cpp
+++ b/test/test_port.cpp
@@ -26,159 +26,212 @@ class PortTest : public PnetIntegrationTest
 
 class PortUnitTest : public PnetUnitTest
 {
+   protected:
+      pnet_t dummystack = {};
 };
 
 TEST_F (PortUnitTest, PortGetPortList)
 {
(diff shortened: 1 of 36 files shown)


olbjo (approved):
Great! I think this is better than previous solution.

test/test_port.cpp
```diff
@@ -30,155 +30,227 @@ class PortUnitTest : public PnetUnitTest
 
 TEST_F (PortUnitTest, PortGetPortList)
 {
-   pnet_t * dummystack = NULL;
+   pnet_t dummystack;
    pf_lldp_port_list_t port_list;
 
    memset (&port_list, 0xff, sizeof (port_list));
+   memset (&dummystack, 0, sizeof (dummystack));
```
hefloryd, line 37: Can we move boilerplate setup code to the text fixture class?

pyhys: Good point.
Moving to `class PortUnitTest`
Done

PR #383, opened 2021-03-17
Improve documentation

pyhys:
Split Codesys page into two pages

Improve logging
Improve error message display in sample app.

PR #384, opened 2021-03-18
Alarm input queue

pyhys:
Based on top of #383

### DIFF
--- a/src/common/pf_alarm.c
+++ b/src/common/pf_alarm.c
@@ -232,62 +232,73 @@ static const char * pf_alarm_apmr_state_to_string (pf_apmr_state_values_t state)
 
 void pf_alarm_show (const pf_ar_t * p_ar)
 {
-   printf ("Alarms   (low)\n");
+   printf ("Alarms\n");
+   printf ("  PNET_MAX_ALARMS             = %d\n", PNET_MAX_ALARMS);
+   printf ("Alarms   (low prio)\n");
    printf (
-      "  alpmi_state         = %s\n",
+      "  alpmi_state                 = %s\n",
       pf_alarm_alpmi_state_to_string (p_ar->alpmx[0].alpmi_state));
    printf (
-      "  alpmr_state         = %s\n",
+      "  alpmr_state                 = %s\n",
       pf_alarm_alpmr_state_to_string (p_ar->alpmx[0].alpmr_state));
    printf (
-      "  apms_state          = %s\n",
+      "  apms_state                  = %s\n",
       pf_alarm_apms_state_to_string (p_ar->apmx[0].apms_state));
    printf (
-      "  apmr_state          = %s\n",
+      "  apmr_state                  = %s\n",
       pf_alarm_apmr_state_to_string (p_ar->apmx[0].apmr_state));
    printf (
-      "  exp_seq_count       = 0x%x\n",
+      "  exp_seq_count               = 0x%x\n",
       (unsigned)p_ar->apmx[0].exp_seq_count);
    printf (
-      "  exp_seq_count_o     = 0x%x\n",
+      "  exp_seq_count_o             = 0x%x\n",
       (unsigned)p_ar->apmx[0].exp_seq_count_o);
    printf (
-      "  send_seq_count      = 0x%x\n",
+      "  send_seq_count              = 0x%x\n",
       (unsigned)p_ar->apmx[0].send_seq_count);
    printf (
-      "  send_seq_count_o    = 0x%x\n",
+      "  send_seq_count_o            = 0x%x\n",
       (unsigned)p_ar->apmx[0].send_seq_count_o);
    printf (
-      "  sequence_number     = %u\n",
+      "  sequence_number             = %u\n",
       (unsigned)p_ar->alpmx[0].sequence_number);
-   printf ("Alarms   (high)\n");
+   printf ("  Incoming alarm queue (mbox) = %p\n", p_ar->apmx[0].p_alarm_q);
    printf (
-      "  alpmi_state         = %s\n",
+      "  Message number in incoming queue = %u\n",
+      p_ar->apmx[0].apmr_msg_nbr);
+
+   printf ("Alarms   (high prio)\n");
+   printf (
+      "  alpmi_state                 = %s\n",
       pf_alarm_alpmi_state_to_string (p_ar->alpmx[1].alpmi_state));
    printf (
-      "  alpmr_state         = %s\n",
+      "  alpmr_state                 = %s\n",
       pf_alarm_alpmr_state_to_string (p_ar->alpmx[1].alpmr_state));
    printf (
-      "  apms_state          = %s\n",
+      "  apms_state                  = %s\n",
       pf_alarm_apms_state_to_string (p_ar->apmx[1].apms_state));
    printf (
-      "  apmr_state          = %s\n",
+      "  apmr_state                  = %s\n",
       pf_alarm_apmr_state_to_string (p_ar->apmx[1].apmr_state));
    printf (
-      "  exp_seq_count       = 0x%x\n",
+      "  exp_seq_count               = 0x%x\n",
       (unsigned)p_ar->apmx[1].exp_seq_count);
    printf (
-      "  exp_seq_count_o     = 0x%x\n",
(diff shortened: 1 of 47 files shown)


src/common/pf_alarm.c
```diff
@@ -978,12 +978,12 @@ static int pf_alarm_apmx_activate (pnet_t * net, pf_ar_t * p_ar)
             p_ar->apmx[ix].frame_id = PF_FRAME_ID_ALARM_HIGH;
          }
 
-         if (p_ar->apmx[ix].p_alarm_q == NULL)
+         /* Input alarm frame queue */
+         if (p_ar->apmx[ix].alarm_receive_q.p_receive_mutex == NULL)
          {
-            p_ar->apmx[ix].p_alarm_q = os_mbox_create (PNET_MAX_ALARMS);
+            p_ar->apmx[ix].alarm_receive_q.p_receive_mutex = os_mutex_create();
```
olbjo, line 984: Maybe "mutex" works as name? p_ should be avoid according to coding standard. That it is for the receive queue is given by the containing struct/variable name

olbjo (commented):
Great!
I focused on the new alarm part but also randomly looked at other parts. Did not review tests

olbjo (approved):
👍

src/common/pf_alarm.c
pyhys: Done

test/test_alarm.cpp
```diff
@@ -294,60 +294,149 @@ TEST_F (AlarmUnitTest, AlarmCheckAddDiagItemToSummary)
...
+
+TEST_F (AlarmUnitTest, AlarmCheckReceiveQueueHandling)
+{
+   pf_alarm_receive_queue_t queue;
+   pf_apmr_msg_t input_frame;
+   pf_apmr_msg_t output_frame;
+   int ix = 0;
+   int err = 0;
+   const uint16_t POSITION_START_NUMBER = 100;
+   const uint64_t POINTER_START_NUMBER = 1000; // TODO uint64_t not portable
```
hefloryd, line 375: Not sure what this is referring to but note that there is a uintptr_t type capable of holding a pointer on the target platform.

src/common/pf_alarm.c
```diff
@@ -38,8 +38,18 @@
  * for high prio.
  *
  * A frame handler is registered for each of the high prio and low prio
- * alarm frame IDs. The incoming frames are put into high prio and low prio
- * queues (mboxes). The periodic task handles frames in the queues.
+ * alarm frame IDs.
+ *
+ * Each AR has two alarm receive queues, one for low prio and one for high
+ * prio alarm frames. The thread responsible for processing incoming frames
+ * is putting frames into the queues, and another thread (driven by the
```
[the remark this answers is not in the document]
pyhys: Done

src/common/pf_alarm.c
```diff
@@ -38,8 +38,18 @@
...
+ * periodic task) is retrieving messages from the receive queues.
+ * A message in an input queue contains a pointer to an incoming frame buffer.
+ *
+ * Each AR has its own alarm send queue. As alarms only can be processed
```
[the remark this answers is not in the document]
pyhys: Done

test/test_alarm.cpp
pyhys: Great, thanks!
Done.

src/common/pf_alarm.c
```diff
@@ -2267,15 +2240,15 @@ static int pf_alarm_send_internal (
    return ret;
 }
 
+/************************ Queue handling ************************************/
```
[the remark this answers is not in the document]
pyhys: Now using same logic for both queues

src/common/pf_alarm.c
```diff
@@ -998,12 +981,10 @@ static int pf_alarm_apmx_activate (pnet_t * net, pf_ar_t * p_ar)
...
          }
 
-         if (p_ar->apmx[ix].p_alarm_q == NULL)
-         {
-            p_ar->apmx[ix].p_alarm_q = os_mbox_create (PNET_MAX_ALARMS);
-         }
-         memset (p_ar->apmx[ix].apmr_msg, 0, sizeof (p_ar->apmx[ix].apmr_msg));
-         p_ar->apmx[ix].apmr_msg_nbr = 0;
+         /* Input alarm frame queue */
+         pf_alarm_queue_mutex_create (
```
hefloryd, line 985: Does the queue need to create/destroy the mutex at runtime? This likely involves malloc/free which is best avoided except for initialization code.

src/common/pf_alarm.c
```diff
@@ -2267,74 +2236,296 @@ static int pf_alarm_send_internal (
...
+
+/**
+ * @internal
+ * Unlock a queue.
+ *
+ * @param p_accountant     InOut: Queue accountant
+ */
+static void pf_alarm_queue_unlock (pf_queue_accountant_t * p_accountant)
+{
+   assert (p_accountant->mutex != NULL);
```
hefloryd, line 2297: CC_ASSERT

src/common/pf_alarm.c
```diff
@@ -2267,74 +2236,296 @@ static int pf_alarm_send_internal (
...
+ * @param p_accountant     InOut: Queue accountant
+ * @param p_read_index     Out:   Read index to use
+ * @return 0 if the queue has an item to read (Queue will be locked)
+ *         -1 if the queue is empty (Will not be locked)
+ */
+static int pf_alarm_queue_lock_and_get_readindex (
+   pf_queue_accountant_t * p_accountant,
+   uint16_t * p_read_index)
+{
+   // TODO Remove, find out why it segfaults sometimes.
```
hefloryd, line 2351: ?

test/test_alarm.cpp
```diff
@@ -294,60 +294,153 @@ TEST_F (AlarmUnitTest, AlarmCheckAddDiagItemToSummary)
...
    int ix = 0;
    int err = 0;
    const uint16_t SEQUENCE_START_NUMBER = 100;
 
    /* Prepare default data */
-   memset (&queue, 0, sizeof (queue));
    memset (&alarm_input, 0, sizeof (alarm_input));
    memset (&alarm_output, 0, sizeof (alarm_output));
-   EXPECT_EQ (queue.count, 0);
+
```
hefloryd, line 309: Check if there is boilerplate setup code here that could be moved to the test fixture.

src/common/pf_alarm.c
```diff
@@ -2267,74 +2236,296 @@ static int pf_alarm_send_internal (
...
+ *         -1 if the queue is full (Will not be locked)
+ */
+static int pf_alarm_queue_lock_and_get_writeindex (
+   pf_queue_accountant_t * p_accountant,
+   uint16_t * p_write_index)
+{
+   os_mutex_lock (p_accountant->mutex);
+   if (p_accountant->count >= PNET_MAX_ALARMS)
+   {
+      os_mutex_unlock (p_accountant->mutex);
```
hefloryd, line 2319: This function does not always exit with the mutex locked. That leads to confusing code later on, where it looks like calling functions forgot to unlock the mutex. Could the code be organised in a way that avoids the "lock_and"? Once you have allocated the write-index, do you need to keep the lock?

src/common/pf_alarm.c
pyhys: The queues are part of the big `pf_ar_t` struct, which is `memset()` on connect request from the PLC.
I think changing this should be a separate PR.

src/common/pf_alarm.c
pyhys: Done

src/common/pf_alarm.c
pyhys: Implemented logic to handle incoming frames before or after the queue is in its initialized state.
Done

test/test_alarm.cpp
pyhys: There are different types for the arguments in the different test cases, so there is not much gain to move to the fixture.
Clarified this by giving better names.
Done.

src/common/pf_alarm.c
pyhys: Restructured the logic.
Done

pyhys:
Renamed queue handling functions to `_post()` and `_fetch()`.

PR #385, opened 2021-03-19
Check IOCS buffer size in CPM

pyhys:


src/common/pf_ppm.h
```diff
@@ -92,6 +92,9 @@ int pf_ppm_set_iocs (
 
 /**
  * Retrieve the data and IOPS for a sub-module.
+ *
+ * This is for reading back the data and the IOPS you have set to the PPM.
```
olbjo, line 96: A bit too personal and a repetition of previous string.

src/common/pf_ppm.c
```diff
@@ -678,10 +681,13 @@ int pf_ppm_set_iocs (
          {
             LOG_ERROR (
                PF_PPM_LOG,
-               "PPM(%d): iocs_len %u expected length %u\n",
+               "PPM(%d): You give IOCS with size %u, but PLC expects size %u "
```
olbjo, line 684: Instead maybe:
"Incorrect IOCS length. Given size %u,  but PLC expects..."
<error message> <error reason>

I think we should avoid using  "you" in log messages.

Applies for all updated log messages.

src/common/pf_ppm.h
```diff
@@ -117,6 +120,9 @@ int pf_ppm_get_data_and_iops (
 
 /**
  * Retrieve IOCS for a sub-module.
+ *
+ * This is for reading back the IOCS you have set to the PPM.
```
olbjo, line 124: Repetition?

olbjo (approved):
👍 

I have some opinions on the actual log messages. We should avoid using "you" and also some more.

src/common/pf_ppm.h
pyhys: Changed to 
`Note that this is not from the PLC, but previously set by the application.`

src/common/pf_ppm.h
pyhys: Changed to 
`Note that this is not from the PLC, but previously set by the application.`
Done

src/common/pf_ppm.c
pyhys: Done

olbjo (approved):
👍

Issue #343, opened 2021-02-15
Linux init() -> deinit()


DominikDessArc:
It might be necessary for our application to restart the driver from within the process without restarting the process.
In the current version of the driver, there is no way to join the threads created by 
'os_thread_t * os_thread_create'
If, for example, I let unit tests using this driver run with --gtest_repeat=xxx, I will have 2 additional threads (1 for eth, 1 for SNMP).
I understand that not having deinit() is not really a problem for microcontrollers since it would never be needed.  
Are there any plans in consideration to add a pnet_deinit() corresponding to pnet_init()?
Am I maybe missing something? So far, I have not found an option that enables me to call pnet_init() again without spawning more and more threads. 
If there is anything that needs clarification, please let me know.
(Development Environment is ubuntu18.04 x86 desktop). The target platform is ubuntu18.04 arm.

pyhys:
Hi, thanks for your input.
Yes, I think that we need to implement this in order to support more use cases. It has not yet been included in the planning, though.

Duplicate of #260

PR #390, opened 2021-03-29
Added destructor for pnet_t members

### LINKED ISSUE #260: Destructor function
Currently, the private function pnet_init_only() is called in the unit tests. This does not allocate the pnet_t struct, but it does allocate memory for mutexes. The allocated memory is never released.

Fix this by creating a destructor function and call it from the tests' TearDown() function.

asade89:
This is a destructor for pnet_init_only(), meaning it does not free the
pnet_t structure itself. It will only free the dynamically allocated members of pnet_t.

Implementing a pnet_init() destructor in addition to this would be
trivial.

Seems like this has a use in tests' TearDown() function to release
dynamically allocated objects.

---

Hello,

I wrote this patch based on discussions in this repo.

This should resolve #260 and potentially #343 (unless `pnet_t` is dynamically allocated in which case it needs to be free'd too).

I was not sure about your standards in naming functions, but I guessed what you would have chosen based on similiar functionality in the code base.

I also noticed you intend to restructure `pnet_t` therefore I kept it simple by freeing the dynamic objects and clearing their pointers. I'm clearing the pointers because that seems to be the norm in the code base.

Introducing a `pnet_init()` destructor is trivial, but I didn't want to expose this to your API as I don't know what your intentions are.

If there are proposals you have in mind, please let me know and I will see can be done.

I have sent a signed CLA by physical mail which should arrive this week :)

Have a good day, cheers :D


pyhys:
Great, thanks for your contribution!
We will have a look.

widavies:
What's the status on this? @pyhys

PR #391, opened 2021-03-30
Clean up pf_block_writer.c

pyhys:
Mock more pnal_xxx functions

Homogenize function prototypes

Make sure that the writer functions are cleaner, and accepting just whe data that should be written.
    
Avoid calls to other files from the block writer functions.


olbjo (approved):
👍

PR #395, opened 2021-03-31
Reintroduce delayed transmission of application ready

pyhys:
Commit 9857aba made each AR have its own CMSU and CMWRR instance, and thus allow
multiple simultaneous connections. To support that the sample app needs to be able
to call pnet_application_ready() for each connection. A (premature) simplification
was made by calling the pnet_application_ready() already in the user callback
for state change, instead of doing a delayed transmission. It did work fine
when testing with connections.

It was later noticed however that it was not possible to do any write after the
connection had been established (normally writes are done during connection setup),
as the CMWRR was left in wrong state.

Reintroduce the delayed transmission of application_ready to solve the problem.

From an end-user point of view it would be nice to be able to call
pnet_application_ready() already from the callback (without using any delay mechanism).
It proved possible to do that by changing the order of how things are done in
the pf_cmdev_state_ind() function. If triggering the user callback last when a state
change occurs, it would be possible as the stack has been properly updated to the
new state. The drawback would instead be that at ABORT conditions, the AR
would already be gone when the user callback is triggered.
Then no error information on what caused the ABORT is available.

One future alternative could be to add an additional argument to the user callback
with error information, instead of asking the AR for error information.

### DIFF
--- a/sample_app/sampleapp_common.c
+++ b/sample_app/sampleapp_common.c
@@ -825,13 +825,11 @@ static int app_state_ind (
          (void)pnet_set_provider_state (net, true);
       }
 
-      if (pnet_application_ready (net, arep) != 0)
-      {
-         printf (
-            "AREP %u Error returned when application telling that it is "
-            "ready for data. Have you set IOCS or IOPS for all subslots?\n",
-            arep);
-      }
+      /* Send application ready at next tick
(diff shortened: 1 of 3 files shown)


olbjo (approved):
👍

PR #400, opened 2021-04-12
Close EPM session after each request

### LINKED ISSUE #396: Endpont mapper (EPM) request does not release session
Multiple EPM requests will use up all available sessions.

pyhys:
TIA portal uses EPM requests for identifying devices on the network.

Previously the sessions were not closed, resulting in "Out of session resources"
when a number of requests have been done.

Profinet uses EPM in a limited way. If we for some reason we need more advanced
EPM handling in the future, implement a timeout for each EPM session.

Closes #396


olbjo (approved):
👍

PR #401, opened 2021-04-13
Allow LLDP and blockwriter functions to be used during stack initiali…

pyhys:
…sation

Previously usage of the LLDP functions during the stack startup caused segfaults,
as the cmdev not was fully initialized. Just changing order of function calls
during startup caused segfaults.


olbjo (approved):
👍

PR #402, opened 2021-04-13
Trigger peer-not-detected diagnosis immediately at startup

pyhys:
If we should verify that a peer is connected, the no-peer-detected
diagnosis should be set already at startup. If the peer shows up,
the diagnosis is removed.

The previous method of setting the diagnosis after some arbitrary time
was prone to intermittent problems, as the boot time varies a lot
on Linux (for example if file system repairs are done at startup).
Dependent on the arbitrary time had timed out or not at the
diagnosis question from the ART Tester, different results appeared.

Found with the test case "Pdev_Check_onePort".


olbjo (approved):
👍

PR #404, opened 2021-04-15
No support for device-access AR

pyhys:
Device-access AR is a limited version of connection as a supervisor can do.
We do not support that.

Update GSDML file to reflect that we support only one concurrent (standard) AR.

In order to support more than one AR in the future, we need to implement
ownership control of different modules in slots.

Found with ART Tester, test case "AR-ASE"


olbjo (approved):
Looks good to me.
Lets wait for tests to complete and proceed if results are good.

PR #405, opened 2021-04-16
Release should consider ARUUID but not the session key

pyhys:
Any session key differences should be ignored, according to
the section 5.2.42.9.2 in Profinet Protocol v 2.4

Found with ART Tester, behavior scenario 8.


olbjo (approved):
👍

PR #406, opened 2021-04-16
Allow only one AR even if PNET_MAX_AR is larger

pyhys:


CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ option (PNET_OPTION_SNMP "" OFF)
 option (PNET_USE_ATOMICS "Enable use of atomic operations (stdatomic.h)" OFF)
 
 set(PNET_MAX_AR                 2
-  CACHE STRING "Number of connections. Must be > 0. 'Automated RT Tester' uses 2")
+  CACHE STRING "Number of connections. Must be > 0. 'Automated RT Tester' uses 2, but only 1 proper AR is supported.")
```
olbjo, line 72: Can we find another word than "proper". Maybe "connection"?

olbjo (approved):
👍 
Minor comment

CMakeLists.txt
pyhys: Done

CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ option (PNET_OPTION_SNMP "" OFF)
 option (PNET_USE_ATOMICS "Enable use of atomic operations (stdatomic.h)" OFF)
 
 set(PNET_MAX_AR                 2
```
hefloryd, line 71: Should the default value be 2? Seems like it would always generate an error (assuming that is what is being checked in pf_cmrpc.c)?

pyhys: Added a comment to describe the need for 2 AR instances in memory.
Done

PR #409, opened 2021-04-21
Base the cyclecounter value on the previous value

pyhys:
Increment the cycle counter value with the known difference.

Previously we based the cyclecounter value on the current time. This fix separates the two
concerns of sending the frames at correct time and to assign proper cyclecounter values.

### DIFF
--- a/src/common/pf_ppm.c
+++ b/src/common/pf_ppm.c
@@ -203,8 +203,8 @@ static void pf_ppm_finish_buffer (
    uint8_t * p_payload = ((pnal_buf_t *)p_ppm->p_send_buffer)->payload;
    uint16_t u16;
 
-   p_ppm->cycle = pf_ppm_calculate_cyclecounter (
-      os_get_current_time_us(),
+   p_ppm->cycle = pf_ppm_calculate_next_cyclecounter (
+      p_ppm->cycle,
       p_ppm->send_clock_factor,
       p_ppm->reduction_ratio);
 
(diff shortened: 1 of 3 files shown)


src/common/pf_ppm.c
```diff
@@ -1023,23 +1025,60 @@ uint16_t pf_ppm_calculate_cyclecounter (
...
+          transmission_interval_in_timebases;
+}
+
+/**
+ * Calculate the next cyclecounter value from the previous one.
+ *
+ * The timebase is 1/32 of a millisecond (31.25 us).
+ *
+ * One cycle length is send_clock_factor x 31.25 us.
+ * The cyclenumber value is keeping track of which cycle we are in now.
```
olbjo, line 1048: cyclenumber / cyclecounter?

src/common/pf_ppm.c
```diff
@@ -1023,23 +1025,60 @@ uint16_t pf_ppm_calculate_cyclecounter (
...
+{
+   uint16_t number_of_previous_transmissions;
+   const uint32_t transmission_interval_in_timebases = send_clock_factor * reduction_ratio;
+
+   /* Integer division for truncation */
+   number_of_previous_transmissions =
+      previous_cyclecounter / transmission_interval_in_timebases;
+
+   /* Now a multiple of transmission intervals */
+   return (number_of_previous_transmissions + 1) * transmission_interval_in_timebases;
```
olbjo, line 1081: If I understand correctly following implementation is easy to follow and gives the same result :

return previous_cyclecounter + send_clock_factor * reduction_ratio;

In my opinion, introducing the transmission_interval_in_timebases intermediate variable just adds complexity.

Maybe I am missing something?

pyhys: This implementation makes sure that we always return a multiple of send_clock_factor * reduction_ratio, regardless of the previous_cyclecounter. See the testcases in this patch.
I will update the documentation.

src/common/pf_ppm.c
pyhys: Removed redundant comment.
Done

olbjo (approved):
👍

PR #410, opened 2021-04-22
Keep UDP socket open at session release

pyhys:
We are missing incoming UDP packets when closing and reopening the UDP socket.
 This was found with the ART Tester, test case Alarm.
 
 The bug is time sensitive, and is triggered when running XMC4800 with debug logging over serial cable at 460800 bit/s.

### DIFF
--- a/src/device/pf_cmrpc.c
+++ b/src/device/pf_cmrpc.c
@@ -4884,11 +4884,8 @@ void pf_cmrpc_periodic (pnet_t * net)
       {
          LOG_DEBUG (
             PF_RPC_LOG,
-            "CMRPC(%d): Closing and reopening socket used for incoming DCE RPC "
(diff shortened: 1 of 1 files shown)


olbjo (approved):
I can't see any problem with this but on the other hand I do not understand the previous close/open. What happens if a packet is stored in the socket when it is being closed?

hefloryd:
> I can't see any problem with this but on the other hand I do not understand the previous close/open. What happens if a packet is stored in the socket when it is being closed?

If the SO_LINGER option is set then close() should block until data is sent or a timeout expires. I didn't find any clear documentation for what happens if SO_LINGER is not set. Transmission may still be queued. SO_LINGER is not used in this project and is also not enabled in LWIP by default.

PR #411, opened 2021-04-23
Increase default number of alarms

pyhys:
In the ART Tester test case "APMS" a burst of 6 alarm frames is sent to the device.
Adapt the default compile time settings so all these incoming frames will fit in the
alarm input queue.


olbjo (approved):
👍

PR #412, opened 2021-04-23
Improve documentation week 16

pyhys:


olbjo (commented):
👍

doc/implementation_details.rst
```diff
@@ -559,6 +577,49 @@ organisation, and the list of supported OIDs in the test case specification
 "Topology discovery check".
 
 
+Dynamic Host Configuration Protocol (DHCP)
+------------------------------------------
+The GSDML file should have the ``AddressAssignment`` attribute set to
+``"DCP;DHCP"``. During the ART Tester test case DCP_OPTIONS_SUBOPTIONS some
```
hefloryd, line 583: Is this always true or just for the ART tester? Presumably there is some use-case where DHCP is not used?

pyhys: Changed to 
```
Dynamic Host Configuration Protocol (DHCP)
------------------------------------------
The GSDML file should have the ``AddressAssignment`` attribute set to
``"DCP;DHCP"`` if DHCP is a supported way to set the IP address.

During the ART Tester test case DCP_OPTIONS_SUBOPTIONS some aspects of DHCP
handling are tested, if the ``AddressAssignment`` attribute is set accordingly.
```
Done

PR #414, opened 2021-04-27
Check number of incoming IOCR blocks

pyhys:
The max number of CR information block we can store is PNET_MAX_CR.
Make sure we do not exceed this.
    
Also do some rudimentary validation of some of the compile time settings.

NOTE: Use the setting "Hide whitespace changes" while reviewing.

### DIFF
--- a/options.h.in
+++ b/options.h.in
@@ -88,8 +88,8 @@
  * These values directly affect the memory usage of the
  * implementation. Sometimes in complicated ways.
  *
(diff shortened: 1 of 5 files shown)


olbjo (approved):
👍 
Additionally maybe a error message should be added?

pyhys:
Added the error message.
Also some compile time setting validation was added.

PR #415, opened 2021-04-27
Log file access times

pyhys:


olbjo (approved):
👍 

Maybe we should print an error or even fatal message if access time is greater than cyclic data period or even tick interval?

pyhys:
Yes, later on we should maybe have some monitoring of the main loop cycle time.

PR #422, opened 2021-05-06
Show details for ports and SNMP

pyhys:


olbjo (approved):

src/common/pf_snmp.c
```diff
@@ -686,3 +686,13 @@ int pf_snmp_get_peer_link_status (
 
    return error;
 }
+
+void pf_snmp_show (pnet_t * net)
+{
+   pf_snmp_data_t * snmp = &net->snmp_data;
+
+   printf ("\nSNMP\n");
+   printf (" SysName     : '%s'\n", snmp->system_name.string);
```
olbjo, line 695: For sample application we have used \"%s\" to indicate string values .
Applies for all string logs added in this commit

olbjo (approved):

src/common/pf_snmp.c
pyhys: Done

PR #425, opened 2021-05-17
Close session after incoming RPC CANCEL, FAULT and REJECT messages.

pyhys:
Set socket value to -1 to indicate that it is closed.
Close sockets only for the sessions initiated by us (for the CControl messaging).
Rename is_release flag to close_socket.


olbjo (approved):
👍 
Looks good to me!

PR #427, opened 2021-05-18
Bugfix RPC max response length

olbjo:
Bug found with ART Tester case "Checking of sending RTC frames by the DUT" for lot of slots with parameter data to be written.


pyhys (approved):
Great!

PR #433, opened 2021-06-18
Allow dcp to respond to identify all filter types

atlowl:
IEC 61158-6-10:2019 4.3.1.2 specifies that an identify filter packet can
contain 1 or more filter blocks and that any block type can be first.
p-net determines that an identify packet is a filter, if the first block is
a station name or alias filter block.

This fixes pf_dcp_identify_req () so that p-net responds to single or
multiple identify filter packets that start any supported filter type.


atlowl:
Closing this, as i did it wrong.

PR #435, opened 2021-06-27
Fix dcp tests

atlowl:
This builds on the previous patch of responding to all filter types. 

The DCP tests were mainly looking for a return of 1 to say that the packet was handled, but pf_dcp_identify_req() would always return 1, the test would therefore always pass.

I believe if the packet isn't right, or if it does not match, then it should not return 'handled'.


test/test_dcp.cpp
```diff
@@ -140,6 +147,19 @@ TEST_F (DcpTest, DcpRunTest)
...
+   ret = pf_eth_recv (mock_os_data.eth_if_handle, net, p_buf);
+   EXPECT_EQ (ret, 1);
+
+   TEST_TRACE ("\nCheck invalid length set ident by device id request fails\n");
+   p_buf = pnal_buf_alloc (PF_FRAME_BUFFER_SIZE);
+   ident_by_device_id_req[32] = 0x01;
+   memcpy (p_buf->payload, ident_by_device_id_req, sizeof (ident_by_device_id_req));
+   ret = pf_eth_recv (mock_os_data.eth_if_handle, net, p_buf);
+   EXPECT_EQ (ret, 0);
+
```
cpackham-atlnz, line 162: I think the check for `mock_os_data.eth_send_count` needs updating below.

atlowl: Correct, this should have returned an extra eth_send_count. Turns out the LLDP packets were being counted as well. Have added a new patch that counts what type of replies were being sent, and now check against that.

PR #437, opened 2021-08-04
Run codespell over code to correct spelling errors

atlowl:
Running codespell to correct spelling mistakes.

Also adding codespell ignore file also for some code related words
that can be safely ignored.

Could split this patch if needed into API documentation / code related.


pyhys:
Thanks for pointing to codespell!
This has been implemented in #464

PR #441, opened 2021-09-17
Build sample app with warnings enabled

hefloryd:
The sample app was being built without warnings. Enable all warnings and fix the resulting issues.


olbjo (approved):
I also saw these warnings when building in another environment and made almost identical updates to fix them.

PR #450, opened 2021-12-08
Check that at least one of the Eth ports is fast enough

pyhys:
Fix error setting mechanism in case of too slow ports

### DIFF
--- a/src/device/pf_pdport.c
+++ b/src/device/pf_pdport.c
@@ -240,6 +240,50 @@ int pf_pdport_save_all (pnet_t * net)
    return 0;
 }
 
+bool pf_pdport_is_a_fast_port_in_use (pnet_t * net)
+{
+   int loc_port_num;
+   pf_port_iterator_t port_iterator;
+   pnal_eth_status_t eth_status;
+   bool found = false;
+   bool fast_enough = false;
+#if LOG_DEBUG_ENABLED(PNET_LOG)
+   const pnet_port_cfg_t * port_config;
+#endif
+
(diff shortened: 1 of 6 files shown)


src/device/pf_pdport.c
```diff
@@ -240,6 +240,50 @@ int pf_pdport_save_all (pnet_t * net)
...
+#if LOG_DEBUG_ENABLED(PNET_LOG)
+   const pnet_port_cfg_t * port_config;
+#endif
+
+   pf_port_init_iterator_over_ports (net, &port_iterator);
+   loc_port_num = pf_port_get_next (&port_iterator);
+   while (loc_port_num != 0)
+   {
+      eth_status = pf_pdport_get_eth_status (net, loc_port_num);
+      fast_enough = eth_status.operational_mau_type >=
```
olbjo, line 259: Maybe only plugged / running interfaces shall be included in check? 
In a multiport scenario I unplugged interfaces/ports could return a default value .

olbjo (commented):
Looks good, one comment to check.

src/device/pf_pdport.c
pyhys: We are looking if there is at least one, so there is no harm in looking at all interfaces (to keep logic simple).
I avoid returning early, just to have the debug log show the speed of all interfaces.

pyhys: Updated to return true if MAU-type is fast enough and the interface is up.

olbjo (approved):
Great!

PR #457, opened 2022-01-11
Use correct argument order for pf_put_padding_align()

pyhys:
According to the documentation for the pf_put_padding_align() function,
the first argument should be the start_position, and the next argument should be the alignment size.

Change the usage to be according to the documentation.

Error reported by user.


pyhys:
Not yet tested

olbjo (approved):

pyhys:
Works fine with test case IP_UDP_RPC_I&M_EPM

pyhys:
Manual checking in Wireshark confirms proper padding

PR #460, opened 2022-01-13
fix warning

SupermanOfHeiLinPu:
Fix when compile generated warning


pyhys:
Hi, and thanks for reporting this.
Which compiler and which settings did you use? 

We typically set the compiler options using target_compile_options() in for example the file cmake/Linux.cmake
Maybe we should update those values.

pyhys:
Closing due to inactivity.

If you have information on which compiler settings to use, please reopen.

PR #463, opened 2022-02-18
Use a separate script to enable GPIOs on Raspberry Pi

pyhys:
The /usr/bin/gpio utility from the wiringpi package is no longer available.
Use a bash script instead.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,6 +4,7 @@
 # Do not perform LF normalization on bash scripts
 *.service text eol=lf
 *.raspberrypi text eol=lf
+*.sh text eol=lf
 src/ports/linux/set_network_parameters text eol=lf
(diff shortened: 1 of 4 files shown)


.gitattributes
```diff
@@ -5,6 +5,7 @@
 *.service text eol=lf
 *.raspberrypi text eol=lf
 src/ports/linux/set_network_parameters text eol=lf
+src/ports/linux/enable-rpi-gpio-pins.sh text eol=lf
```
hefloryd, line 8: *.sh is usually eol=lf

olbjo (approved):
:+1:

.gitattributes
pyhys: Changing all .sh to eol=lf

PR #468, opened 2022-03-01
Use sh shell instead of bash for Linux scripts

pyhys:


hefloryd:
Use .sh suffix and set gitattributes using *.sh?

pyhys:
The stack uses `set_network_parameters`. If I remember correctly we did not specify a script file ending, as users might use for example Python for these settings.

PR #469, opened 2022-03-02
Centralise porthandling

pyhys:


src/common/pf_eth.c
```diff
@@ -158,6 +158,23 @@ int pf_eth_init (pnet_t * net, const pnet_cfg_t * p_cfg)
...
 }
 
+int pf_eth_send_on_physical_port (
+   pnet_t * net,
+   int loc_port_num,
+   pnal_buf_t * buf)
+{
+   pf_port_t * p_port_data = pf_port_get_state (net, loc_port_num);
+
+   // TODO Handle tail tagging: Put loc_port_num in p_buffer metadata
```
olbjo, line 168: I think you should use a /*  */ comment

olbjo (approved):
This works for me. Checkin commit mentions that this is a preparation for tail-tagging which is good.
:+1:

src/common/pf_eth.c
pyhys: Done

olbjo (commented):
Looks good!

PR #470, opened 2022-03-04
Better handling of Ethernet MAU types

pyhys:
When the link is down the MAU type can not be read from
hardware, for Linux and possibly other operating systems.

Add a configuration value for default MAU type value, to be used
when the actual value not is available.


olbjo (approved):
Nice!

PR #474, opened 2022-03-09
Join profinet multicast group (linux)

olbjo:
Some linux solutions needs this addition to receive
Profinet multicasts. DCP Identity all do not work
without it.

Change-Id: Ia7fb34d1c23e1ebcacda2dd76964c2ae488584c3

### DIFF
--- a/src/ports/linux/pnal_eth.c
+++ b/src/ports/linux/pnal_eth.c
@@ -20,6 +20,10 @@
 
 #include "pnal.h"
 
+#include "pnet_options.h"
+#include "options.h"
+#include "osal_log.h"
+
 #include <net/ethernet.h>
 #include <net/if.h>
 #include <netpacket/packet.h>
@@ -92,6 +96,8 @@ pnal_eth_handle_t * pnal_eth_init (
    struct sockaddr_ll sll;
    int ifindex;
    struct timeval timeout;
+   struct packet_mreq mreq;
+   const uint8_t pn_mcast_addr[ETH_ALEN] = {0x01, 0x0e, 0xcf, 0x00, 0x00, 0x00};
    const uint16_t linux_receive_type =
       (receive_type == PNAL_ETHTYPE_ALL) ? ETH_P_ALL : receive_type;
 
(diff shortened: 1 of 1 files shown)


src/ports/linux/pnal_eth.c
```diff
@@ -142,6 +143,24 @@ pnal_eth_handle_t * pnal_eth_init (
    sll.sll_protocol = htons (linux_receive_type);
    bind (handle->socket, (struct sockaddr *)&sll, sizeof (sll));
 
+   /* Join profinet multicast group */
+   mreq.mr_ifindex = ifindex;
+   mreq.mr_type = PACKET_HOST | PACKET_MR_MULTICAST;
+   mreq.mr_alen = ETH_ALEN;
+   const uint8_t pn_mcast_addr[ETH_ALEN] = {0x01, 0x0e, 0xcf, 0x00, 0x00, 0x00};
```
pyhys, line 150: Vi brukar lägga definitioner i början av funktionen

pyhys (commented):
Ser bra ut. En minimal justering.

src/ports/linux/pnal_eth.c
olbjo: OK

pyhys (approved):
Sweet!

src/ports/linux/pnal_eth.c
```diff
@@ -142,6 +144,23 @@ pnal_eth_handle_t * pnal_eth_init (
...
+
+   if (
+      setsockopt (
+         handle->socket,
+         SOL_PACKET,
+         PACKET_ADD_MEMBERSHIP,
+         &mreq,
+         sizeof (mreq)) != 0)
+   {
+      printf ("Failed to join Profinet multicast group\n");
```
hefloryd, line 161: This should probably be a log message of some sort. Should the error be propagated to the caller?

olbjo: I added the log message as a PNAL message. All pnal files share the same tag.
Regarding the message I just wanted to keep this problem visible and then address it if it is a problem on other systems. 

Return codes for other ioctls or options are not checked in this function. It has worked without this additional socket option with previous linux systems we have been using and I do not think we need to propagate the error to user. 

I do not know why additional socket option is needed in this particular case. Just tried make it work.

PR #481, opened 2022-03-24
Error codes from write operations should be sent back to the IO-controller

pyhys:
Now tested with ART tester


pyhys:
Tested on ART tester (test cases without device B)

olbjo (approved):
:+1: 
Verkar vettigt

PR #485, opened 2022-04-05
Show a log message if the stack refuses to change IP address

pyhys:
Simplify logic in pf_cmina_dcp_set_commit() by returning early.

Increase log level to INFO.


olbjo (approved):
Ser bra ut!

PR #486, opened 2022-04-08
Set detailed cause when reporting errors from pf_cmrpc_rm_connect_int…

pyhys:
…erpret_ind()

This will show the detailed error cause in the message sent back to the PLC,
which simplifies Wireshark debugging.

The description of pf_cmrpc_rm_connect_interpret_ind() says that the rpc_result field
should be updated at error. This fix makes sure that the rpc_result is set for all error cases.


olbjo (approved):
Kanonbra!

PR #495, opened 2022-05-04
rt-kernel: update patch for 2021.1

### LINKED ISSUE #476: rt-kernel patch does not apply cleanly on 2021.1
lwIP in rt-kernel 2021.1 seems to have changed enough to make the patch non-trivial to port to the newer kernel tree.

hefloryd:
This updates the p-net patch for rt-kernel for the 2021.1 release.

Fix #476.


pyhys (approved):

nattgris:
I've seen some cases where I've ran out of pbufs with these changes on an STM32F4 (dwmac1000 driver) -> all Ethernet frames discarded -> device completely unresponsive via Ethernet. Perhaps some pbuf leak at some point? Seems to be after several reloads of a webpage with several resources loaded in parallell from an web server on the device. Will investigate a bit more.

PR #506, opened 2022-06-03
Improve scheduler

pyhys:


olbjo (commented):
I had a major comment on adding  current time to the api. Think it though and comment on how you want to do it. 

After that I can do some additional review of the scheduler internals

PR #510, opened 2022-07-01
fix NULL dereference

### LINKED ISSUE #475: NULL pointer dereference in rt-kernel pnal_get_hostname
If the hostname is not defined in the rt-kernel config (CFG_LWIP_HOSTNAME), the rt-kernel port will strcpy from a NULL pointer (netif_default->hostname).

hefloryd:
pnal_get_hostname for rt-kernel and STM32Cube could dereference a NULL
pointer if lwip had not been configured with a hostname. Check before
using hostname and return error if no hostname has been set.

Fix #475


pyhys (approved):
Great!

PR #516, opened 2022-08-09
Fix reported media type when eth link is down

olbjo:
Use correct default media type for unconnected
port. Problem seen in test case Different Access
Ways Port 2 Port when run with LAN9962 setup.


olbjo:
Hold on. Maybe default_mau_type should be used instead.

olbjo:
This is instead fixed by updating the app utils for the lan9662 sample

PR #517, opened 2022-08-10
Limit the DCP identify to a single response

pyhys:
An matching incoming DCP identify request will schedule a delayed response.
Limit to a single waiting response, in order not to run out of scheduler resources
and not to overwrite the already waiting response.


pyhys:
Not yet tested

pyhys:
Replaced by #528

PR #520, opened 2022-08-25
Fix default route with multiple interfaces

bwachtendorf:
The old script overwrote the default route. This worked on systems with one network interface.
But this lead to breaking the internet connection when multiple interfaces are present.
This change sets the route only for that network.


pyhys:
I think we should avoid setting the default gateway on Linux.
See #561

pyhys:
Thanks for your contribution. See #561

PR #531, opened 2023-02-03
Add support for shared device

amn-rtl:
Having support for shared device is at its core the same as having support for more than one AR. This commit adds such support.

The main changes are:
- Added handling more than one concurrent AR.
- Better separation of real and expected identifications.
- Added submodule ownership handling.
- Change the PNET_MAX_AR configuration setting to be the actual number of application relations.

Change-Id: Ie8792cf66317a834391430a7a785e5702f69fe22


pyhys (approved):
Great!
You have already fixed my previous comments.

PR #534, opened 2023-02-06
Modify the Github workflow

amn-rtl:
Add workflow_dispatch, and use apt-spy2 to find Ubuntu mirrors.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,5 @@
 name: Build
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 env:
   BUILD_TYPE: Release
 jobs:


hefloryd:
I am not familiar with apt-spy2 but seems like it may have it's own issues, see https://github.com/actions/runner-images/issues/675#issuecomment-612446737. Any particular reason for using Swedish mirrors? 

The problem with the azure servers seem to be intermittent and may fix itself, although of course a robust permanent fix would be nice.

amn-rtl:
I am not familiar with apt-spy2, either. It was suggested as a fix for the problematic azure servers, which I have been waiting for most of today. I chose Swedish mirrors mostly because the --launchpad option to apt-spy2 required a specific country, and the --launchpad option was suggested to be somewhat more safe than letting the program pick from a more general mirror list. Indeed, when I ran without the option, a mirror that did not have all that we needed was picked.

I agree that a permanent fix would be needed.

amn-rtl:
I am prepared to drop this. But, since the checks for a PR might fail because of unavailable servers, and there is no way to manually retrigger the test, I assume that such a failure does not stop a PR from being merged.

hefloryd:
Yes, with sufficient privileges it should be possible to merge a PR even if the build fails. My preferences would be to close this for now and reopen it if azure acts up again.

pyhys (approved):
Great!

amn-rtl:
I have removed the search for alternative mirrors, but kept the workflow_dispatch event trigger. Could we try if that helps us with manually rebuilding?

PR #545, opened 2023-04-17
Two protocol fixes

amn-rtl:
Here are two fixes that add some sanity checks and their respective failure responses to the controller.


pyhys (approved):
Great

olbjo (approved):
:+1:

olbjo (approved):
:+1:

PR #546, opened 2023-04-20
Correct the accounting of outgoing RPC fragment sizes

amn-rtl:
For outgoing fragmented RPC messages, the header size was accounted for an extra time for each fragment except the first one. This change corrects that.


pyhys (approved):
This is about receiving an ACK on a fragment we have sent earlier, and the calculation of how much data we have successfully sent so far. For the first ACK we should count the entire message, but for consecutive messages we should only count the payload (not the header).

PR #559, opened 2023-06-09
Script to clean a directory from alarm-free wireshark files

pyhys:


olbjo (approved):
👍 
I realize this may be useful in some scenarios. 
Is there any section in the documentation that is pointing
to this script or should a reference from the documentation 
be added?

PR #561, opened 2023-06-22
Avoid setting default gateway on Linux

pyhys:


pyhys:
Replaces #520

olbjo (approved):
👍 
Possibly the commit message could describe the 
problem this commit is fixing.

pyhys:
Updated the commit message

PR #572, opened 2024-01-15
Expand MAU Type Support for 2.5G, 5G and 10G 

ahmkar94:


ahmkar94:
Sorry, managed to set the wrong upstream for my PR -- this will be reported later via Addiva Elektronik

PR #581, opened 2024-05-08
Fix pnal_buf_free does nothing on non debug builds

dns13:
As asserts only are evaluated on debug builds their calls need to be free of side effects. In this case we leak all pbufs as they were never freed.


dns13:
Sorry, already fixed.

